PR #1, opened 2019-11-18
Remove wm-specific methods. Rework tests.

danielballan:
This PR removes the function ``format_version`` and the ``WaybackClient``
methods ``timestamped_uri_to_version`` and ``list_versions`` as discussed
in https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/477.

It reworks the tests to exercise the methods ``search`` and ``get_memento``
directly.

Also, when splitting this package off from web-monitoring-processing, I wrongly
omitted the VCR artifacts (cassettes) from version control. They should be in
version control, and they are reinstated here.


danielballan:
Closes https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/477

PR #3, opened 2019-11-18
Port 3b024e20 from web-monitoring-processing.

danielballan:


wayback/_client.py
```diff
@@ -492,7 +495,12 @@ def search(self, url, *, matchType=None, limit=None, offset=None,
                 if isinstance(value, str):
                     final_query[key] = value
                 elif isinstance(value, datetime):
-                    final_query[key] = value.strftime(URL_DATE_FORMAT)
+                    # Make sure we have either a naive datetime (assumed to
+                    # represent UTC) or convert the datetime to UTC.
+                    value_utc = value
+                    if value.tzinfo:
+                        value_utc = value.astimezone(tzutc())
```
Mr0grog: I *think* we can just use `value.astimezone(datetime.timezone.utc)` here, right? That way we don’t need the extra `dateutil` dependency. (Guess I should have done that in the original.)

danielballan: Verified interactively and pushed.

danielballan:
Added a test as well and rebased on `master`. Mind taking a quick look at the test to ensure it is sensible (timezones are confuuuuusing), @Mr0grog? Merge away, if so.

Issue #4, opened 2019-11-18
Expose and document CdxRecord


Mr0grog:
`WaybackClient.search()` yields instances of `CdxRecord`, which makes it a public API: https://github.com/edgi-govdata-archiving/wayback/blob/800608fc06a154c5a75442762e0984d0cb72c000/wayback/_client.py#L101-L114

We should at least document it, and probably (?) also expose it publicly.

**I don’t think this is required for the 0.2 release.**

danielballan:
Closed by #5

PR #5, opened 2019-11-19
Make CdxRecord public and document it.

danielballan:
Note TODO regarding the documentation for ``CdxRecord.key``.

This also prompts the question, should we convert `length` and `status_code`
to integers? The `length` field in particular seems odd as a string.

### DIFF
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -112,6 +112,61 @@ CdxRecord = namedtuple('CdxRecord', (
     'raw_url',
     'view_url'
 ))
+"""
+Item from iterable of results returned by :meth:`WaybackClient.search`
+
+These attributes contain information provided directly by CDX.
+
+.. py:attribute:: digest
+
+   Hash of captured content given as a string, such as
+   :data:`'MGIGF4GRGGF5GKV6VNCBAXOE3OR5BTZC'`.
+
+.. py:attribute:: key
+
+   TODO: Something related to SURT?
+
+.. py:attribute:: length
+
+   Size of captured content in bytes, given as a string such as :data:`'2767'`.
+
+.. py:attribute:: mime_type
+
+   MIME type of record, such as :data:`'text/html'`.
+
+.. py:attribute:: status_code
+
+   Status code returned by the server when the record was captured, given as a
+   string such as :data:`'200'`.
+
+.. py:attribute:: timestamp
+
+   The capture time represented as a string of the form :data:`'YYmmddHHMMSS'`.
+   See the attribute :attr:`date` below for a more convenient representation.
+
+.. py:attribute:: url
+
+   The URL that was captured by this record, such as
+   :data:`'http://www.nasa.gov/'`.
+
+And these attributes are synthesized from the information provided by CDX.
+
+.. py:attribute:: date
+
+   The capture time represented as a :class:`datetime.datetime`, such as
+   :data:`datetime.datetime(1996, 12, 31, 23, 58, 47)`.
+
+.. py:attribute:: raw_url
+
+   The URL to the raw captured content, such as
+   :data:`'http://web.archive.org/web/19961231235847id_/http://www.nasa.gov/'`.
+
+.. py:attribute:: view_url
+
+   The URL to the public view on Wayback Machine, surrounds the captured
+   content with a navigation panel, such as
+   :data:`'http://web.archive.org/web/19961231235847/http://www.nasa.gov/'`.
+"""
 
 
 def split_memento_url(memento_url):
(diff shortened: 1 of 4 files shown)


danielballan:
Closes #4

wayback/_client.py
```diff
@@ -112,6 +112,61 @@ def __init__(self, retries, total_time, causal_error):
...
 ))
+"""
+Item from iterable of results returned by :meth:`WaybackClient.search`
+
+These attributes contain information provided directly by CDX.
+
+.. py:attribute:: digest
+
+   Hash of captured content given as a string, such as
+   :data:`'MGIGF4GRGGF5GKV6VNCBAXOE3OR5BTZC'`.
```
Mr0grog: Should note this is a base 32 encoded SHA-1.

wayback/_client.py
```diff
@@ -112,6 +112,61 @@ def __init__(self, retries, total_time, causal_error):
...
+
+.. py:attribute:: key
+
+   TODO: Something related to SURT?
```
Mr0grog: Yep, this is the SURT-formatted URL.

wayback/_client.py
```diff
@@ -112,6 +112,61 @@ def __init__(self, retries, total_time, causal_error):
...
+
+.. py:attribute:: length
+
+   Size of captured content in bytes, given as a string such as :data:`'2767'`.
```
Mr0grog: Note: may be inaccurate if a revisit record (looks like it’s the length of the reference, not the revisited response).

wayback/_client.py
```diff
@@ -112,6 +112,61 @@ def __init__(self, retries, total_time, causal_error):
...
+
+.. py:attribute:: mime_type
+
+   MIME type of record, such as :data:`'text/html'`.
```
Mr0grog: May be `'warc/revisit'` if a revisit record, even though useful mime type info is contained elsewhere. May also be `'unk'` if this data wasn’t captured for some reason.

wayback/_client.py
```diff
@@ -112,6 +112,61 @@ def __init__(self, retries, total_time, causal_error):
...
+
+.. py:attribute:: status_code
+
+   Status code returned by the server when the record was captured, given as a
+   string such as :data:`'200'`.
```
Mr0grog: May be `'-'` if a revisit record.

wayback/_client.py
```diff
@@ -112,6 +112,61 @@ def __init__(self, retries, total_time, causal_error):
...
+.. py:attribute:: raw_url
+
+   The URL to the raw captured content, such as
+   :data:`'http://web.archive.org/web/19961231235847id_/http://www.nasa.gov/'`.
+
+.. py:attribute:: view_url
+
+   The URL to the public view on Wayback Machine, surrounds the captured
+   content with a navigation panel, such as
+   :data:`'http://web.archive.org/web/19961231235847/http://www.nasa.gov/'`.
```
Mr0grog: Also updates the URLs for links and subresources in the document to Wayback URLs, if HTML, JS, or CSS. Not just the added nav panel that’s different.

wayback/_client.py
```diff
@@ -525,7 +587,16 @@ def search(self, url, *, matchType=None, limit=None, offset=None,
...
-                data = CdxRecord(*text.split(' '), None, '', '')
+                data = CdxRecord(*text.split(' '), '', '')
+                if data.status_code == '-':
+                    # the status code given for a revisit record
+                    status_code = None
+                else:
+                    try:
+                        status_code = int(data.status_code)
+                    except TypeError as err:
+                        raise UnrecognizedStatusCode(status_code) from err
```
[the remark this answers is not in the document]
danielballan: The custom exception may be overzealous, but it will inject some potentially-useful specificity into the traceback. Commit dda82af19cd0e8fba4a1b6dd975afdc25f1b4434 makes the chaining explicit. The final exception will still be `UnexpectedResponseFormat`, which is accurate, but when applicable it will be clear that the direct cause was `UnrecognizedStatusCode`. I don't think we need custom exceptions for _every_ case. This just struck me as the riskiest bit of parsing that we do; I am hedging against the possibility that CDX might throw _other_ strings beyond `'-'` at us.

danielballan:
Rebased.

wayback/_client.py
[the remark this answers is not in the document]
danielballan: I would also be totally fine with removing this extra exception in the name of code simplicity. Feel free to ask for that revision or push a commit of your own at will to expedite. :-)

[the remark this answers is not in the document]
Mr0grog: Ah, good point, I wasn’t thinking about the exception chaining aspect.

I don’t think I’m as convinced as you that status codes are a more special case, but I also don’t really feel strongly about it. Totally OK to merge as-is, I think! 👍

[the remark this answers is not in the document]
danielballan: Upon reflection, it's not worth a special exception type. Removed. Merging....

PR #6, opened 2019-11-19
Make original_url_for_memento public.

danielballan:


wayback/__init__.py
```diff
@@ -2,4 +2,7 @@
 __version__ = get_versions()['version']
 del get_versions
 
-from ._client import WaybackClient, WaybackSession  # noqa
+from ._client import (  # noqa
+    original_url_for_memento,
```
Mr0grog: Obviously this is the one I was using, but I’m wondering whether we should expose `memento_url_data` alongside it or instead (since it gets *all* the useful info out of the URL).

danielballan: :+1: for instead.

danielballan:
OK. You convinced me that adding an attribute is _also_ a good idea, but there is no harm in exposing this, so I will merge in the name of progress.

PR #10, opened 2019-12-02
Remove vestigial bits from `_utils` module

Mr0grog:
We had a whole bunch of vestigial bits in the `_utils` module that were left over from web-monitoring-processing but not used here. This removes them and also removes the `lxml` dependency, which was only present because of a vestigial utility function.

While I was at it, I also moved the `SessionClosedError` class to the `exceptions` module, since I completely missed it in #7.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,2 @@
-lxml
 requests
 urllib3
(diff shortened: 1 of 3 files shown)


Mr0grog:
> Oh, the lxml dependency should have tipped me off that there was more to excise from this package.

That’s exactly what did it for me! :P

PR #21, opened 2020-02-05
Make sure to close responses where they may leak

Mr0grog:
This is a bit of a follow-on to #19 and #20.

These are all the remaining places I could find possible connection leaks from responses being held onto. A lot of these cases *should* be covered by the requests library (it takes a similar approach and caches response bodies before raising most exceptions), but I didn't feel like I could be certain that was always true, so I went with the most cautious approach: anywhere we aren't explicitly returning an open request, read/cache the body and close it.

### DIFF
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -215,6 +215,17 @@ def cdx_hash(content):
     return b32encode(hashlib.sha1(content).digest()).decode()
 
 
+def read_and_close(response):
+    # Read content so it gets cached and close the response so
+    # we can release the connection for reuse. See:
+    # https://github.com/psf/requests/blob/eedd67462819f8dbf8c1c32e77f9070606605231/requests/sessions.py#L160-L163
+    try:
+        response.content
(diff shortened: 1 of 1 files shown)


wayback/_client.py
```diff
@@ -215,6 +215,17 @@ def cdx_hash(content):
...
 
+def read_and_close(response):
+    # Read content so it gets cached and close the response so
+    # we can release the connection for reuse. See:
+    # https://github.com/psf/requests/blob/eedd67462819f8dbf8c1c32e77f9070606605231/requests/sessions.py#L160-L163
+    try:
+        response.content
+    except (ChunkedEncodingError, ContentDecodingError, RuntimeError):
+        response.raw.read(decode_content=False)
+    response.close()
```
danielballan: Far be it from me to try to improve on an implementation copied straight out of `requests`, but I might move this inside a `finally:` block in case any other unexpected Exception types are raised above.

Mr0grog: Oh, nice thinking! 👍

Issue #2, opened 2019-11-18
Wrap result from `get_memento` in a custom object


danielballan:
Rather than exposing `requests.Response` directly, we could wrap it in a custom object to give us the freedom to refactor in the future. From https://github.com/edgi-govdata-archiving/web-monitoring-processing/issues/477

Mr0grog:
In Web Monitoring, we are adding better media type parsing (see edgi-govdata-archiving/web-monitoring-processing#621), and that should probably ultimately live here instead. I’m thinking we’d have:

```python
class MediaType:
    type: str        # e.g. 'text'
    subtype: str     # e.g. 'html'
    parameters: str  # e.g. {'charset', 'utf-8'}

    @property
    def media(self):
        "The main media type without parameters, e.g. 'text/html'"
        return f'{self.type}/{self.subtype}'

	@property
    def parameter_string(self):
        "The parameters as a single string, e.g. 'charset=utf-8; other-param=whatever'"
        return '; '.join(f'{key}={value}' for key, value in self.parameters)

    def __str__(self):
        if parameters:
            return f'{self.media}; {self.parameter_string}'
        else:
            return self.media
```

And then `Memento.media` would be an instance of the above.

Mr0grog:
Or it could be simpler, with just:

```python
memento.media_type == 'text/html'
memento.media_parameters == {'charset': 'utf-8'}
```

Mr0grog:
OK, thinking through an ideal model for a Memento, here’s my first cut:

- `encoding` The text encoding of the response. (Do we need an [`apparent_encoding` like Requests has](https://requests.readthedocs.io/en/latest/api/#requests.Response.apparent_encoding) that detects the encoding if one isn’t specified in the headers?)

- `content` The content of the response in bytes.

- `text` The decoded content of the response as a string.

- `headers` The *archived* headers of the response (that is, the headers that came with the original snapshot, not including additional Wayback headers).

- `history` A list of mementos for any *historical* redirects that led to this memento.

- `debug_history` A list of URLs for any redirects (including those that are involved in general communication with the Wayback Machine, not just historical redirects). (Should this be a bunch of response objects of some sort instead? Memento objects that aren’t actually mementos?)

- `is_redirect` Whether the memento is of a redirect (i.e. had a 3xx status code).

- `ok` Whether the status code was < 400.

- `status_code` The status code of the response as an integer.

[...]

PR #23, opened 2020-02-07
Sketch out a way to support multithreading

Mr0grog:
**This is less an actual PR than a sketch of one possible way to do things. It’s not necessarily meant to merge (or at least not exactly as-is).** Trying to figure out how to take what I did in edgi-govdata-archiving/web-monitoring-processing#551 and integrate it more smoothly here.

This implementation is kinda wonky, but is the best way I've come up with to support sessions/clients across multiple threads and pooling connections across multiple threads. It's based on the kind of hacky implementation in edgi-govdata-archiving/web-monitoring-processing#551.

The basic idea here includes two-and-a-half pieces, and works around the fact that urllib3 is thread-safe, while requests is not:

1. `WaybackSession` is renamed to `UnsafeWaybackSession` (denoting it should only be used on a single thread) and a new `WaybackSession` class just acts as a proxy to multiple UnsafeWaybackSessions, one per thread.

2. A special subclass of requests's `HTTPAdapter` that takes an instance of urllib3's `PoolManager` to wrap. `HTTPAdapter` itself is really just a wrapper around `PoolManager`, but it always creates a new one. This version just wraps whatever one is given to it.

3. `UnsafeWaybackSession` now takes a `PoolManager` as an argument, which, if provided, is passed to its `HTTPAdapter`. `WaybackSession` creates one `PoolManager` which it sets on all the actual `UnsafeWaybackSession` objects it creates and proxies access to. That way a single pool of requests is shared across many threads.

This is super wonky! It definitely makes me feel like we might just be better off dropping requests and just using urllib3 directly (especially given #2 -- which means requests wouldn't be part of our public interface in any way). But this is a smaller change that *probably* carries less short-term risk.


wayback/_client.py
```diff
@@ -387,8 +402,73 @@ def reset(self):
...
+        concrete_session = self.get_unsafe_session()
+        return concrete_session.__enter__()
+
+    def __exit__(self, *args):
+        concrete_session = self.get_unsafe_session()
+        return concrete_session.__exit__(*args)
+
+    def request(self, *args, **kwargs):
+        concrete_session = self.get_unsafe_session()
+        return concrete_session.request(*args, **kwargs)
```
Mr0grog: Almost certainly there’s a better way to handle this (e.g. with magic methods, or maybe [wrapt](https://wrapt.readthedocs.io/)).

Mr0grog:
Ok, given the lack of feedback, I am probably going to move forward with testing, cleaning up, and merging this approach this week. I don’t think it’s a *good* approach, though, so I’m also going to start moving forward with #2 and switching to using urllib3 directly instead of requests as a next step. (Httpx claims to be headed for an API-stable release by April, so I want to look into that, too.)

Mr0grog:
This has been sitting for a long time! A general update: the more I stare at this, the less comfortable I feel with it. It introduces a super-wonky abstraction to get around Requests’s lack of thread-safety and, while that makes multi-threaded use much easier, it probably introduces some maddening corner cases.

I’m pretty sure there’s no emergency here (nobody’s complained, and EDGI [a.k.a me] is the only user who desperately needs thread safety, but they already have a workaround, even if it’s ugly), so it’s probably better to take the time to put together a nicer solution. I think that means implementing #2 so that Requests is no longer part of our API, and then swapping Requests out for either urllib3 or Httpx, both of which should be thread-safe. (I’m less clear on Httpx’s sync mode being thread safe, so it needs more investigation. It’s also still in beta.)

I’m leaving this open in case the above doesn’t work out or we run into a real need to ship multithreading support sooner, but I don’t currently expect to finish and merge it.

PR #24, opened 2020-02-13
Add final release notes for v0.2.2

Mr0grog:
@danielballan do these read OK? I feel like the connection fixes we implemented are working well, and demand a release at this point. (The crazy proposal in #23 seems big enough and speculative enough to be worth a separate release.)

If this looks good, I was thinking I’d cut a release straightaway (Thursday, Feb. 13th).

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -2,12 +2,17 @@
 Release History
 ===============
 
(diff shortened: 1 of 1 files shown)


Mr0grog:
Anchors aweigh! ⚓️

PR #26, opened 2020-03-21
Allow from_date and to_date to be a datetime.date

### LINKED ISSUE #25: Searching with datetime.date instead of datetime.datetime
I noticed that when supplying a `from_date` and `to_date` as a `datetime.date` instead of a `datetime.datetime` that the search results are not time-boxed correctly. This is because a `datetime.date` doesn't get formatted correctly when it is sent to the CDX API.

edsu:
This PR adds a test and new functionality in wayback.WaybackClient to allow users to pass in a `datetime.date` for `from_date` and `to_date`. I think the parameter names suggest that a `datetime.date` would suffice.

When `WaybackClient.search` is given a `datetime.date` as either I think either an exception should be thrown, or it should allow for them. My gut tells me that the latter would be better, and that is what is included in this pull request. But I leave it up to you to decide whether being permissive in this case is the right approach. No hard feelings if you think strict is the way to go.

Thanks for a very useful module!

Fixes #25


edsu:
Yep, just commented on another PR that I would be grateful to be listed.

PR #28, opened 2020-03-21
Support Python 3.6

### LINKED ISSUE #27: Python 3.6?
I was trying to use wayback in a [Colab Jupyter Notebook](https://research.google.com/colaboratory). It wouldn't install because wayback specifies 3.7 in its setup.py and Colab is using python v3.6. 

I tested wayback under python3.6 and it seemed to work fine, so I was wondering if the version in setup.py could be relaxed a little bit?

edsu:
This commit relaxes the setup.py to allow for python 3.6, which still passes the test suite. I also took a shot at updating the circleci setup to test 3.6, but I'm not 100% sure I did that correctly or that it works.

Fixes #27

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  build:
+  test-3.7: &test-template
     working_directory: ~/wayback
     docker:
       - image: circleci/python:3.7
@@ -45,8 +45,14 @@ jobs:
             . ~/venv/bin/activate
             make -C docs html
 
+  test-3.6:
+    <<: test-template
+    docker:
+      - image: circleci/python:3.6
+
 workflows:
   version: 2
   build:
     jobs:
-      - build
+      - test-3.6
+      - test-3.7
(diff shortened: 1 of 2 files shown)


danielballan:
Thanks, @edsu.

I cannot recall our original reasoning for choosing 3.7 as the minimum supported version. Likely it’s just the lowest version we needed for our application. Seems reasonable that we should support 3.6 as wel at this time..

Given that the tests pass, I’m inclined to merge and release once @Mr0grog has taken a look at the Circle-CI config.

Mr0grog:
I think we inherited v3.7 when extracting this from [web-monitoring-processing](https://github.com/edgi-govdata-archiving/web-monitoring-processing), but the only thing over there that actually requires it is our use of `SimpleQueue`. We don’t use `SimpleQueue` here, though, so it seems safe to me. Let me get Circle building for this, though. (I thought we had it set to build forked PRs, but clearly not.)

edsu:
Of course, I'd love to be listed as I'm a big fan of the EDGI project and Web Monitoring effort.

PR #30, opened 2020-03-24
Add contributors and copyright info to the README

Mr0grog:
I tried to use something resembling our standard all-contributors format, but couldn’t for the life of me figure out how to do it in reStructuredText. There doesn’t seem to be any obvious way to add a `title` attribute to a link, and tables with emoji in them wouldn’t render to begin with.

I debated just switching to Markdown, but that didn’t seem worthwhile just for a more nicely formatted contributors list, so I settled for this.

Also added licensing info while I was at it.


Mr0grog:
This also removes the `AUTHORS.rst` file, so the way we address contributors here is more in keeping with other projects across EDGI.

PR #45, opened 2020-06-20
Raise `RateLimitError` for 429 responses

### LINKED ISSUE #22: Add custom error for rate limit issues?
During some recent testing, I ran into some [HTTP 429 (too many requests)](https://httpstatuses.com/429) responses in `WaybackClient.search()` calls. At the moment, we wrap *any* HTTP error in search in a generic `WaybackException`: https://github.com/edgi-govdata-archiving/wayback/blob/b502236b8069e6731f5803d509e84248a1ab7448/wayback/_client.py#L565-L567

But since 429 responses often come with a `Retry-After` header, it might be useful to have a custom error for this case. Or at least make sure the response object is included as an attribute on the error.

Mr0grog:
Status code 429 ("Too Many Requests") specifies that you've hit a rate limit, and the Wayback Machine makes good use of it. We now raise a special error type for this case since it's something a user might want to have special handling for.

Fixes #22.

### DIFF
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -37,7 +37,8 @@ from .exceptions import (WaybackException,
                          UnexpectedResponseFormat,
                          BlockedByRobotsError,
                          MementoPlaybackError,
-                         WaybackRetryError)
+                         WaybackRetryError,
+                         RateLimitError)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


wayback/_client.py
```diff
@@ -336,6 +337,8 @@ def send(self, *args, **kwargs):
             try:
                 result = super().send(*args, **kwargs)
                 if retries >= maximum or not self.should_retry(result):
+                    if result.status_code == 429:
+                        raise RateLimitError(result)
```
Mr0grog, line 341: Worth noting here: we raise the error *after* exhausting retries. It’s probably debatable whether that’s correct; I figured that, since the retry wait might be long enough to automatically overcome the error, this was OK.

PR #50, opened 2020-09-21
Change `url`, add `datetime` & `mode` arguments to `get_memento`

### LINKED ISSUE #15: Support original URL + date or CdxRecord instances as parameters for `get_memento`?
At current, `get_memento()` requires the actual URL of a memento as its main parameter:

```py
wayback_client.get_memento('https://web.archive.org/web/20190513000000/https://www.epa.gov/')
```

That’s fine for EDGI’s existing use cases because we always query CDX for a list of mementos to retrieve. However, it seems reasonable that people would also want to get the memento for a given web page nearest to a given time:

```py
wayback_client.get_memento('https://www.epa.gov/',
                           date=datetime(2019, 5, 13),
                           type='raw',
                           exact=False)
```

It might also be a nice convenience to accept `CdxRecord` instances:

```py
wayback_client.get_memento(some_cdx_record,
                           type='raw',
                           exact=False)
```

Some questions to solve here:

- Would these be good to add?

- Should they all be ways to call `get_memento` or do we need separate methods for [some] of them?

- When getting a memento via these new approaches, we need some additional info: the type of memento to return. In memento URLs, you can add the following strings to the end of the timestamp to control the memento type:

    - `` (nothing) gets you a memento where all the subresource URLs are replaced with URLs for mementos (so you can reliably render it in a browser, for example) and that has extra UI from Wayback (custom header, timeline, etc).
    - `id_` gets you the raw memento with no changes.
    - `js_` gets you JavaScript with some extra comments indicating debug and copyright info.
    - `cs_` gets you CSS with some extra comments indicating debug and copyright info.
	- `im_` gets you an image, presumably with some kind of extras like the above (otherwise you’d just use `id_`). Not actually sure on the details here.
    - Only docs I’ve ever been able to find for this feature: http://archive-access.sourceforge.net/projects/wayback/administrator_manual.html#Archival_URL_Replay_Mode

    In the examples above, I’ve mocked this up as `type='raw'` (indicating the `id_` type). Lots of options here as to how we treat this (just use the Wayback strings above, have some constants, use custom names like in my example, etc). For EDGI’s case, we always want `id_`, and I suspect that’s a reasonable default, but seems like it ought to be an option since I can certainly imagine other use cases.

- Is `exact=True` still the right default? If requesting an memento URL (as we currently do) or a `CdxRecord` instance, that makes sense, but maybe not if you are asking for original URL + datetime.

/cc @danielballan

Mr0grog:
Fixes #15. The basic idea here is that the current interface to `get_memento`, where you must pass an actual *memento URL* rather than the URL of the page you want a memento for, is pretty unintuitive for almost every case except EDGI's current usage. In other words, instead of doing:

```py
client.get_memento('https://web.archive.org/web/20200101000000id_/http://www.epa.gov')
```

It's often more sensible to do:

```py
client.get_memento('http://www.epa.gov', date_=date(2020, 1, 1))
```

You don’t need special knowledge of how Wayback Machine URLs are constructed, and it’s much easier to read. We still support the old behavior above (both for backwards compatibility and because it's useful).

In addition, since our main use case boils down to:

```py
for cdx_record in client.search('<some url>'):
    client.get_memento(cdx_record.raw_url)
```

We may as well allow the CDX record itself, too:

```py
for cdx_record in client.search('<some url>'):
    client.get_memento(cdx_record)
```

---

### Concerns/Things to Think Through

- [x] `date_` is a super awkward parameter name. I will probably change our `datetime.date` import in the module to `from datetime import date as Date` to resolve this so we can use `date`.

    However! @edsu pointed out in #26 that the name “date” is maybe not ideal in the first place, since we are *really* talking about a particular time (but a date is acceptable — we treat is a low resolution date-time). I chose `date_` for now to have consistency with the rest of the functions we have, but I’m wondering if this should be something else, like `time`, `when`, etc.

- [x] I’ve tacked `date_` on at the end for compatibility, but since the idealized standard usage of this method is now:

    ```py
    client.get_memento('https://example.com/whatever', date_=date(2020,1,1))
    ```

    It seems like this really should be the second parameter so you aren’t forced to name it in simple cases:

    ```py
    client.get_memento('https://example.com/whatever', date(2020,1,1))
    ```

    That would be a breaking change, but we are still at v0.x for a reason. 😉 Should I do that and release this as v0.3.0 instead of v0.2.5?

    Relatedly, it seems like a good breaking change to force everything else to be keywords, so the new signature would be:

    ```py
    def get_memento(url, date_=None, mode='id', *, exact=True,
                    exact_redirects=None, target_window=24 * 60 * 60
                    follow_redirects=True):
    ```

- [x] In #15, I proposed the name `type`, but in this PR, it’s `mode`. We *can* use `type` (it’s not reserved), but it is a built-in global, and shadowing it seems non-ideal. Does `mode` sound OK? My biggest concern is that it conflicts with the usage of “mode” in [the only documentation I know of for this feature](http://archive-access.sourceforge.net/projects/wayback/administrator_manual.html#Archival_URL_Replay_Mode) — “mode” there is about how the whole server is operating, not the way it’s playing back a particular memento, and this feature is is essentially unnamed. (It’s “a flag” on the URL, but maybe that’s OK and we should call it `flag`? Feels too generic to me, though.)

- [x] `mode` takes string values that aren’t defined anywhere as a constant/enum. Should they be? It also uses the the names from the Wayback API (`''`, `'id_'`, `'cs_'`, etc.) instead of custom names, although it lets you drop the ugly trailing underscore. I think that’s the right decision (bad to have too many special things that don’t match up the actual standardized API), although it leaves `''` as somewhat unclear (plus, you’d normally think the empty string would be the default, but it’s not here).

- [x] Making `mode` a changeable parameter means I could probably clean up some of the other private functions a bit more as part of this PR. (e.g. we now have `ARCHIVE_URL_TEMPLATE`, `ARCHIVE_RAW_URL_TEMPLATE`, and `ARCHIVE_VIEW_URL_TEMPLATE`, but they can all be handled by just `ARCHIVE_URL_TEMPLATE`.)

- [x] Need to verify the docs and update the tutorial.


danielballan:
Agreed that `date_` is awkward. I usually only reach for that naming pattern if the collision would involve a keyword or (maybe) a builtin. Seems like `datetime` which accepts instances of `datetime` or `date`.

I like the proposed new signature, relegating the more "settings"-oriented kwargs to keyword-only. I think it's worth doing and moving to v0.3.0 over.

I read over that section of the docs trying to find inspiration for a better name. I can't think of anything better than `mode`. I agree that `flag` is not good. I _do_ think it might be worth making an `enum` to make the set of allowed values specified in a machine-readable way. It would also provide a space for more descriptive names, taken from the Wayback docs. The function signature could tolerate both an `enum` member and its string equivalent, for convenience.

```py
In [6]: class Mode(enum.Enum):
   ...:     original = 'id_'
   ...:     formatted = ''  # I just made this up. Maybe we can do better than 'formatted'.
   ...:     javscript ='js_'
   ...:     css = 'css_'
   ...:     image = 'im_'
   ...: 

In [7]: Mode(Mode.original)
Out[7]: <Mode.original: 'id_'>

In [8]: Mode('id_')
Out[8]: <Mode.original: 'id_'>
```

However, this cuts against:

>  I think that’s the right decision (bad to have too many special things that don’t match up the actual standardized API), although it leaves '' as somewhat unclear (plus, you’d normally think the empty string would be the default, but it’s not here).

so I can see the argument for sticking with strings. I lean slightly toward enums. :man_shrugging:

Mr0grog:
> I can see the argument for sticking with strings. I lean slightly toward enums.

I think that makes sense. Enum names that are more friendly with values that match the API seem like a reasonable way to split the difference on this kind of thing.

Re: `date_`, 

> Agreed that `date_` is awkward. I usually only reach for that naming pattern if the collision would involve a keyword or (maybe) a builtin. Seems like `datetime` which accepts instances of `datetime` or `date`.

Just to make sure I’m following: are you suggesting `datetime` would be a better name, or that `date` is clear enough about what you should be putting in there? (Obviously agreed the underscore should go.)

Mr0grog:
Rebased this on `main` since #49 merged.

Frijol:
fwiw re

> the name “date” is maybe not ideal in the first place, since we are really talking about a particular time (but a date is acceptable — we treat is a low resolution date-time)

I think `datetime` is more intuitive, I would be surprised if a `date` took time but unsurprised if `datetime` accepted just a date

docs/source/release-history.rst
```diff
@@ -7,6 +7,40 @@ In Development
...
 [Add information about changes in your PR here]
 
+**Breaking Changes:**
+
+- The order of parameters in :meth:`wayback.WaybackClient.get_memento` has changed, and additional parameters for the date and mode of memento playback have been added. You can now call ``get_memento()`` with the URL and time you want the memento of, rather than needing to compose a complete Wayback memento URL. That is, you can now call:
+
+  .. code-block:: python
+
+     client.get_memento('http://www.noaa.gov/',
+                        datetime(2018, 8, 16, 11, 19, 11))
```
Frijol, line 17: you still need to know something about date formatting in order to know what's going on here, it would be helpful to see a line like "where datetime's arguments are (4-digit year, month number, day of month, ...)"

Frijol: or even e.g. "uses standard UTC" would be a nice breadcrumb

docs/source/release-history.rst
```diff
@@ -7,6 +7,40 @@ In Development
 
 [Add information about changes in your PR here]
 
+**Breaking Changes:**
+
+- The order of parameters in :meth:`wayback.WaybackClient.get_memento` has changed, and additional parameters for the date and mode of memento playback have been added. You can now call ``get_memento()`` with the URL and time you want the memento of, rather than needing to compose a complete Wayback memento URL. That is, you can now call:
```
Frijol, line 12: might be helpful to call out "can" vs "must", specifically

> We still support the old behavior above (both for backwards compatibility and because it's useful).

Mr0grog: I’m not sure I’m following here. Which advice/guidance needs clarification?

docs/source/release-history.rst
Mr0grog: Do you really think how to use `datetime` needs to be spelled out here? It’s part of Python’s standard library and widely used all over the place.

Re: UTC, good point, although `datetime` can carry timezone information, and if it does, we respect it. So it’s more that we assume UTC if you don’t specify a timezone. Anyway, I’ll add something to that effect.

FWIW, I don’t necessarily expect anything here to be the final release notes. Since this is going to be a breaking change, I’m going to include #2 in this release, and between this PR and that, we’re pretty much totally redo-ing everything about `get_memento`. So it’s going to need much more explanation that might involve rewriting a bunch of this.

Frijol: no, you're right. This is fine

docs/source/release-history.rst
Frijol: Listing this change under "breaking changes" makes it read as though it no longer supports the old format – but my read of your description elsewhere is that the old format is still supported. Is that correct? Or are the parameters moved around even in the old format?

Mr0grog: Ahhhh, so if you used to use it in the most simple way, you’re fine and nothing breaks:

```py
client.get_memento('http://web.archive.org/web/20180816111911id_/http://www.noaa.gov/')
```

And if you used any of the other parameters as keyword arguments instead of positional arguments, you’re still fine:

```py
client.get_memento('http://web.archive.org/web/20180816111911id_/http://www.noaa.gov/',
                   exact=False,
                   exact_redirects=False,
                   target_window=3600)
```

BUT you used to be able to specify all those other things as positional arguments, too, which is no longer supported. So this used to work, but now breaks:

```py
# Same as above, but using positional instead of keyword args. No longer works.
client.get_memento('http://web.archive.org/web/20180816111911id_/http://www.noaa.gov/',
                   False,
                   False,
                   3600)
```

I should probably break up the discussion of the arguments that are now keyword-only (the breaking part) from the *new* arguments (`mode`, `date_`, and new values for `url`). That would be clearer.

Mr0grog: (I could have not made the above breaking change, but the code would be messier, and we’re about to break more stuff with #2, so this seemed like a good time to be breaking things in the name of code organization. 😄)

docs/source/release-history.rst
Mr0grog: Cool. I will still update to clarify that it gets interpreted as UTC when you don’t explicitly specify a timezone. That’s a good catch! 👍

Mr0grog:
@danielballan re: the no value mode, do any of these jump out at you as especially good?

- `formatted`
- `reformatted`
- `browser`
- `browsable`
- `view` (We’re using this term for the same purpose in `CdxRecord`, although that doesn’t mean it was an especially well-considered choice.)
- `viewable`
- `archive`
- `archived`
- `rewrite`
- `rewritten`

danielballan:
Sorry for the lag, just noticed the notifications about your follow-up questions.

> Just to make sure I’m following: are you suggesting datetime would be a better name, or that date is clear enough about what you should be putting in there? 

Yes, I meant that `datetime` would be a good parameter name. I think shadowing core module names with parameter names is completely fine.

Regarding the enum member name, I think a lot of those are good. My favorites are:

* `reformatted`
* `view`

Mr0grog:
Augh, I am the odd one out who doesn't like `datetime`! (Agree `date` is not great, though.)

Re: mode enum, from those remaining options, I'll go with `view`. I like `reformatted` a lot, but the JS and CSS modes are also reformatting things, `view` says a little more about the intent.

🙌

Mr0grog:
OK! I think this is ready to land. I’m going to come back to it with fresh eyes tomorrow before merging. It’s grown pretty large with all the added docs and tests at this point, so I won’t expect additional review.

@danielballan & @Frijol let me know if you plan to re-review and I’ll hold off on merging.

PR #52, opened 2020-10-02
Return a new `Memento` type from `get_memento()`

Mr0grog:
Rather than returning a requests.Response object from `WaybackClient.get_memento()`, we now return a completely new `Memento` type. It’s similar to requests.Response in a lot of ways, but it contains additional properties that are specific to mementos (like a timestamp) and has different values that reflect the historical archive rather than the exact response the Wayback Machine sent (e.g. `url` is the URL of the archived page, not `http://web.archive.org/web/<date>/<url>`; `headers` contains only the historic headers, not additional info added by the Wayback Machine).

There are two major reasons for this:

1. We can provide some more useful conveniences like the above noted bits.

2. We can change the underlying HTTP implementation, since we no longer expose it publicly. This is critical to getting thread safety, since Requests is not thread safe and it doesn’t look like that’s going to change soon or really ever be something they want to guarantee. Actually swapping out the underlying HTTP library is not intended to be part of this PR, though. That’ll be done separately.

    (It’ll be sad to lose some of Requests’s conveniences, but we aren’t making general purpose HTTP calls, so many of them aren’t as necessary to us — we have a lot more constraints around our usage.)

Note this drops a few features from #2 that I ultimately decided were a bit too complex (e.g. detailed media type parsing) and/or non-critical (e.g. parsing out all the timemap info from the `Links` header). We can add those in later.

**To-do:**

- [x] Code reorganization. I kinda feel like there was already too much in `_client.py` and this really pushes it well beyond.
- [x] Decide on `Memento.timestamp` vs. `Memento.datetime`.
- [x] Validate this API against scripts in web-monitoring-processing to make sure we aren’t leaving any majorly helpful features out or making anything too complicated.

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -74,6 +74,50 @@ In Development
 
 - :func:`wayback.memento_url_data` now returns 3 values instead of 2. The last value is a string representing the playback mode (see above description of the new ``mode`` parameter on :meth:`wayback.WaybackClient.get_memento` for more about playback modes).
 
+- :meth:`wayback.WaybackClient.get_memento` now returns a :class:`wayback.Memento` instance instead of a ``requests.Response`` object. It has a similar interface, but provides more useful information about the Memento than just the HTTP response from Wayback. For example, ``memento.url`` is the original URL the memento is a capture of (e.g. ``http://www.noaa.gov/``) rather than the Wayback URL (e.g. ``http://web.archive.org/web/20180816111911id_/http://www.noaa.gov/``). You can still get the full Wayback URL from ``memento.memento_url``.
+
+  You can check out the full API docs for :class:`wayback.Memento`, but here’s a quick guide to what’s available:
+
+  .. code-block:: python
+
+     memento = client.get_memento('http://web.archive.org/web/20180816111911id_/http://www.noaa.gov/home',
+                                  exact=False)
+
+     # These values were previously not available except by parsing
+     # `memento.url`. The old `memento.url` is now `memento.memento_url`.
+     memento.url == 'http://www.noaa.gov/'
+     memento.timestamp == datetime.datetime(2018, 8, 29, 8, 8, 49, tzinfo=timezone.utc)
+     memento.mode == 'id_'
+
+     # Used to be `memento.url`:
+     memento.memento_url == 'http://web.archive.org/web/20180816111911id_/http://www.noaa.gov/'
+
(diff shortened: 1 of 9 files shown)


Mr0grog:
Marking this as ready to go! I think this is working reasonably well now and I’m relatively happy with it. Over in web-monitoring-processing, the [`test-wayback-0.3` branch](https://github.com/edgi-govdata-archiving/web-monitoring-processing/compare/main...test-wayback-0.3) makes use of it, and seems to work great.

Mr0grog:
Rebased to merge cleanly after porting #53 over to the `main` branch.

Mr0grog:
I think this is basically ready to go now. I took a stab at rewriting the release notes in a more final form here, combining all the changes to `get_memento()` into a slightly clearer, condensed writeup. Check the preview: https://wayback--52.org.readthedocs.build/en/52/release-history.html#in-development

After this lands, I’ll publish an alpha release `v0.3.0a1` so folks can play with it a bit before cutting a final release with such a big change.

docs/source/release-history.rst
```diff
@@ -9,75 +9,142 @@ In Development
...
-     Don’t worry, though — complete Wayback URLs are still supported. This code still works fine:
+  Finally, you can now specify the *playback mode* of a memento using the ``mode`` parameter:
 
-     .. code-block:: python
+  .. code-block:: python
 
-        client.get_memento('http://web.archive.org/web/20180801000000id_/http://www.noaa.gov/')
+     client.get_memento('http://www.noaa.gov/',
+                        datetime=datetime.datetime(2018, 8, 1),
+                        mode=wayback.Mode.view)
```
danielballan, line 81: I am happy with the way this approached turned out.

wayback/_models.py
```diff
@@ -0,0 +1,287 @@
...
+   Content hashed as a base 32 encoded SHA-1.
+
+.. py:attribute:: key
+
+   SURT-formatted URL
+
+.. py:attribute:: length
+
+   Size of captured content in bytes, such as :data:`2767`. This may be
+   innacurate. If the record is a "revisit record", indicated by MIME type
```
danielballan, line 34:
```suggestion
   inaccurate. If the record is a "revisit record", indicated by MIME type
```

danielballan (approved):
One minor copyedit. Looks great.

Mr0grog:
:shipit:  :shipit:  :shipit: / ⚓ AWEIGH!

PR #53, opened 2020-10-19
Fix `target_window` when redirect is > 1 day

Mr0grog:
The `target_window` parameter for `get_memento()` is supposed to limit how far off in time from the requested time you can get a memento for when the `exact` parameter is `False`. However, it only works correctly when the target is off by less than a day! (The scenarios we were originally concerned with in EDGI almost invariably were on the scale of a few hours, so I guess that's how this error snuck in. 😬) If you set `exact=True` (the default), this bug wouldn’t be triggered.

We were checking the target offset by the number of seconds, irrespective of the additional number of days involved. We should have been checking by `total_seconds`, which converts the days into seconds and includes them. I’ve also added tests here to make sure this problem doesn’t occur again.

This PR targets the v0.2.x branch; I’ll add a new PR for `main` (which is now for v0.3.x) after this merges.

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -5,7 +5,7 @@ Release History
 v0.2.5 (2020-10-19)
 -------------------
 
-This release fixes a bug where the ``target_window`` parameter for :meth:`wayback.WaybackClient.get_memento` did not work correctly if the memento you were redirected to was off by more than a day. See `#53 <https://github.com/edgi-govdata-archiving/wayback/pull/53>`_ for more.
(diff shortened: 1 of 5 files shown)


Mr0grog:
Example case — If you set a ``target_window`` of 5 minutes:

```py
url = 'https://web.archive.org/web/20200101000000id_/https://epa.gov/'
memento = client.get_memento(url, exact=False, target_window=5 * 60)
```

This would correctly fail if the resulting memento was off by 10 minutes, but incorrectly succeed if the memento was off by 1 day and 4 minutes.

Issue #66, opened 2021-03-01
Sessions should have a default timeout


Mr0grog:
In https://github.com/edgi-govdata-archiving/web-monitoring-processing, we got through a lot of work to carefully configure sessions for downloading Mementos, but have a bug I’m not fairly sure is caused by not having timeouts on searches (it is possible for for the CDX API to hang for… a *long* time).

I think we need to pick some reasonable default timeout on [`WaybackSession`](https://github.com/edgi-govdata-archiving/wayback/blob/7e9bccc8113e2a24090d3c51b1ae463871da8181/wayback/_client.py#L229) to help avoid this issue. Not yet sure what good values would be.

LionSzl:
Setting a timeout in the Session did not solve the problem, since (from what I understand) it is never applied.

WaybackSession.send sets the timeout that is passed on to the send call to the session's timeout, only if both the session's timeout is not None and the kwargs do not already contain a timeout. But when no timeout is specified in the original call, the kwargs still contains a timeout key with None as value, therefore the sessions timeout does not overwrite this.

This snippet shows the problem:
```python
import requests
class CustomSend(requests.Session):

    def __init__(self):
        super().__init__()

    def send(self, *args, **kwargs):
        print(kwargs)
        super().send(*args, **kwargs)

session = CustomSend()
res = session.request('GET', 'https://duckduckgo.com')
```
> {'timeout': None, 'allow_redirects': True, 'verify': True, 'proxies': OrderedDict(), 'stream': False, 'cert': None}

I've fixed the issue here: https://github.com/LionSzl/wayback/commit/46ee8c60b20d269c33b4ad0f7b7498f77fb580e9 Should I make a pull request?

Mr0grog:
Whoa, good catch!

A PR would be great, BUT I’m not sure about that particular solution, since it prevents someone from sending a request with no timeout if for some reason they need to. I can think of two approaches:

1. Allow `timeout=0` or `timeout=(x, y)` (where x or y could be 0) and change values <= 0 to `None`.

2. Also override `WaybackSession.request()` and add the logic there, too.

Neither seems great (departing in a weird way from Requests’ API vs. doubled-up, redundant logic), but both solve the issue. I'd lean towards (2) since it doesn't change how the API is *supposed* to work. If you have another, better idea, feel free to use it in your PR.

I also still think it makes sense to set a default.

LionSzl:
[...]

PR #68, opened 2021-03-15
Ensure session timeout is used, add default

LionSzl:
The bug described in #66, therefore the session now respects the default ratelimit, if no other ratelimit was provided when calling `WaybackSession.request`.

Default timeout of 60 seconds.


wayback/_client.py
```diff
@@ -226,7 +226,7 @@ class WaybackSession(_utils.DisableAfterCloseSession, requests.Session):
     # just the error type. See `should_retry_error()`.
     handleable_errors = (ConnectionError,) + retryable_errors
 
-    def __init__(self, retries=6, backoff=2, timeout=None, user_agent=None):
+    def __init__(self, retries=6, backoff=2, timeout=60, user_agent=None):
```
Mr0grog, line 229: Ah, this is actually really long! (Have you encountered many Wayback cases where this is what you need?)

This isn’t a timeout for the whole request/response cycle, but a timeout between reads on the network socket — i.e. this is saying if you go more than N seconds between receiving packets from the Wayback Machine servers, then cancel. Usually this is more like 2 (short, but common) to 20 (pretty long for most use-cases). Since Wayback is definitely a little slow, I think something more on the long end is good, but 60 seems pretty far out there to me. 😳

Mr0grog (commented):
> However, for now there is no way to actually provide a timeout value to the requests function without hacking the code.

I’m not sure I’m following. Do you have an example of what you mean? It seems like someone could still do:

```py
session = wayback.WaybackSession()
# Custom timeout:
session.request('http://something.com/', timeout=5)
# or no timeout:
session.request('http://something.com/', timeout=None)
```

wayback/_client.py
Mr0grog: Actually, maybe it’s fine that this is long. It’s at least some outside bound! Nevermind my earlier comment.

Mr0grog:
Also, could you please add tests for this?

wayback/_client.py
LionSzl: Right, the timeout is accessible from within the package. What I meant is that a user cannot provide a custom timeout when calling `WaybackClient.search`. But I don't know if this is even necessary.

Yeah, 60 seconds is extremely long. But I've had problems with my script completely freezing every couple hundred searches. So this timeout would prevent the me from waiting days for a response without the chance of stopping an request that is actually getting answered.

LionSzl:
> Also, could you please add tests for this?

Sure. I've not yet worked with pytest so give me a couple days.

wayback/_client.py
Mr0grog: > What I meant is that a user cannot provide a custom timeout when calling `WaybackClient.search`. But I don't know if this is even necessary.

Ah, I see! I don’t think we need to change that in this PR (the current path for doing that is creating a custom session when you instantiate the client), but it would be great to add `timeout` as a keyword argument to both `search()` and `get_memento()` as a follow-on to this. Can you make an issue for that?

> I've had problems with my script completely freezing every couple hundred searches. So this timeout would prevent the me from waiting days for a response without the chance of stopping an request that is actually getting answered.

👍 👍 👍

LionSzl: OK

LionSzl:
So, I've added the the tests.
It's somewhat hacky, I've had no idea how to properly check if the timeouts were applied. So I patched the `requests.Session.send` method to return the provided timeout value back to the caller. If you know of a better solution, let me know and I'll implement it.

I've forgotten to add the mock library to the dev dependencies, that is what the third commit was for.

wayback/tests/test_client.py
```diff
@@ -2,6 +2,8 @@
 from pathlib import Path
 import pytest
 import vcr
+import requests
+import mock
```
Mr0grog, line 6: Since this project requires Python 3.6+, you can just use the built-in `unittest.mock` library rather than adding a new dependency. `mock` is a backport of the built-in for older versions of Python.

wayback/tests/test_client.py
```diff
@@ -474,6 +476,14 @@ def test_get_memento_follow_redirects_does_not_follow_historical_redirects():
         assert len(memento.debug_history) == 1
 
 
+def return_timeout(self, *args, **kwargs) -> requests.Response:
+    """Overwirte the requests.Session.send to return a response with the provied timeout value as content"""
```
Mr0grog, line 480: Typo:

```suggestion
    """Overwrite requests.Session.send to return a response with the provided timeout value as content."""
```

wayback/tests/test_client.py
```diff
@@ -514,3 +524,37 @@ def test_rate_limit_error_includes_retry_after(self, requests_mock):
...
+    def test_timeout_applied_session(self, mock_class):
+        # Is the timeout applied through the WaybackSession
+        session = WaybackSession(timeout=1)
+        res = session.request('GET', 'http://test.com')
+        assert res.text == '1'
+        # Overwriting the default in the requests method
+        res = session.request('GET', 'http://test.com', timeout=None)
+        assert res.text == 'None'
+        res = session.request('GET', 'http://test.com', timeout=2)
+        assert res.text == '2'
```
Mr0grog, line 538: This *is* hacky, but I think it works fine for now! I’m fine merging this. 👍

If you want to do something fancier, you could check out this timeout test in web-monitoring-diff: https://github.com/edgi-govdata-archiving/web-monitoring-diff/blob/2fe3088b04bb2a58fdb5fec1e92d4fed181233f3/web_monitoring_diff/tests/test_server_exc_handling.py#L192-L202

(It tests that a proxy server correctly times out while contacting the “upstream” server with the actual content. The `patch` decorator makes the proxy being tested have a short timeout, and `responder` and `SimpleHttpServer` create a quick little HTTP server to use as the upstream, which waits longer than the expected timeout before responding. That’s way more complex than what you’re doing here, though. It would be more complete, but not really necessary. Also note that package uses `async/await`, so the solution here couldn’t be exactly the same.)

Mr0grog (commented):
This looks great. I noted a few things inline. The big one is just that we shouldn’t need to install `mock`, which is just a backport of the builtin `unittest.mock`, which we should be able to use instead.

wayback/tests/test_client.py
LionSzl: Huh, I didn't know that mock came build in with unittest.
I've changed that (and the typos below).

Mr0grog:
Is this good to go? If so, I’ll merge later tonight and backport to the v0.2.x release line. (And then publish v0.2.6.)

LionSzl:
From my side it is good to go.

Mr0grog:
@LionSzl I should have asked this before: would you like to be added to [the contributors list in the README](https://github.com/edgi-govdata-archiving/wayback#contributors)?

LionSzl:
> @LionSzl I should have asked this before: would you like to be added to [the contributors list in the README](https://github.com/edgi-govdata-archiving/wayback#contributors)?

Sure, if one bugfix qualifies me as contributor... My full name is Lion Szlagowski.
Thanks!

Mr0grog:
Of course it does! And you are now in the README. 😄

Mr0grog:
I’ve also backported the bugfix part of this and published it as v0.2.6.

Issue #12, opened 2019-12-02
Rethink rate limiting


Mr0grog:
`WaybackClient.get_memento` has left-over rate-limiting behavior from web-monitoring-processing: https://github.com/edgi-govdata-archiving/wayback/blob/f1cdb1dc2692b4215f00a2c05c53fe9c733b24f9/wayback/_client.py#L645-L648

From the perspective of this more generic module…

- Should there be rate limiting at all?
- If there should be, it should probably be optional and/or configurable: the rate shouldn’t be hard-coded, and should maybe be an option on the session or client constructor.
- If there should be, should it apply across both `get_memento` and `search` (and whatever other methods `WaybackClient` might gain in the future)? Or maybe it should apply at the level of making a request to Wayback, rather than at the higher-level `get_memento` method?

/cc @danielballan 

---

## Updates

- We should keep rate limiting.
- It should be expanded to cover `search`.
- `search` and `get_memento` should fall under separate rate limits that don’t interfere with each other.
- The rate limit should be configurable (and should continue to apply across all client/session instances on all threads). **The right API for this is not yet clear.**
- The default limit for CDX calls should be 1/second and for Memento calls should be 30/second.

Mr0grog:
I guess we could also supply it as a separate helper (rather than building it into the client or session) for people to use or ignore as they see fit. There might also be better generic tools for this out there and we should just drop this [possibly too-simple] internal implementation. Not sure. ¯\\\_(ツ)\_/¯

danielballan:
There is an argument for continuing to include simple built-in rate limiting as a default so that we guide our users toward good manners. Users coming from the analysis side of things might not know they should do this, or how to do it effectively. Since this library lends itself toward scraping data in batch from IA, the potential for accidental misuse seems significant.

As a first step, I suggest we apply our simple rate limiting consistently and clearly document how to opt out and leave rate limiting up to the caller. If we discover a generic tool for rate limiting that we are willing to rely on as dependency, great—no need to make a special one here—but IIRC your initial investigation didn’t uncover one.

Mr0grog:
> There is an argument for continuing to include simple built-in rate limiting as a default so that we guide our users toward good manners.

👍

> If we discover a generic tool for rate limiting that we are willing to rely on as dependency, great—no need to make a special one here—but IIRC your initial investigation didn’t uncover one.

[...]

PR #79, opened 2021-03-22
Ratelimits

LionSzl:
I've written a first suggestion of the rate limits, mentioned in #12. It expands the rate limit to `WaybackClient.search` and allow the limits to be set through the `WaybackSession`.
Furthermore, these limits can be set (e.g. disabled for one call) through a function argument.

Should there also an option to change the limits session wide?


wayback/_client.py
```diff
@@ -611,7 +626,8 @@ def search(self, url, *, match_type=None, limit=1000, offset=None,
 
     def get_memento(self, url, datetime=None, mode=Mode.original, *,
                     exact=True, exact_redirects=None,
-                    target_window=24 * 60 * 60, follow_redirects=True):
+                    target_window=24 * 60 * 60, follow_redirects=True,
+                    calls_per_second=None):
```
Mr0grog, line 630: @LionSzl same as with `search()`: do you have a use-case that requires configuration on each call, rather than differently configured clients? I’d prefer not to have this in order to keep things simpler.

wayback/tests/test_client.py
```diff
@@ -623,3 +624,75 @@ def test_timeout_empty(self, mock_class):
...
         res = session.request('GET', 'http://test.com', timeout=1)
         assert res.text == '1'
+
+    @ia_vcr.use_cassette()
+    def test_search_rate_limits(self):
+        # The timing relies on the cassettes being present,
+        # therefore the first run might fail.
+        # Since another test might run before this one,
+        # I have to wait one second before starting.
+        time.sleep(1)
```
Mr0grog, line 634: Since you aren’t using the fancier locking class I mentioned in the issue (https://github.com/edgi-govdata-archiving/wayback/issues/12#issuecomment-787589273), I’m thinking a good solution here is to add an instance-specific prefix to the `group` name on the `_utils.rate_limited()` calls, e.g:

```py
with _utils.rate_limited(calls_per_second=calls_per_second if calls_per_second is not None else
                                          self.session.memento_calls_per_second,
                         group=f'{id(self.session)}:get_memento'):
    # rate-limited code here
```

Mr0grog (commented):
@LionSzl I’m sorry this wound up sitting unaddressed for so long! I’ve rebased this and cleaned it up to merge smoothly with the latest code on `main`.

I’m hoping to get it merged into v0.4.0 this week. This might be off your radar now, so if I don’t hear back from you after a few days, I’ll go ahead and make some more changes here in order to merge.

Mr0grog:
FYI the docs build failure is caused by Scipy’s docs site being down, not an issue in the code here: https://github.com/edgi-govdata-archiving/wayback/pull/101#pullrequestreview-1162480695

wayback/tests/test_client.py
Mr0grog: Hmmm, actually, I’m not sure I should do this without having a thread-safe client/session. The current approach forces all sessions to share the same timings across threads (that is, they all see the same “last time called” when determining how long to wait), which is actually a good thing if you’ve got multiple threads working in parallel, each with one client. We don’t want that situation to accidentally cause you to exceed Wayback’s rate limits.

On the flip side, this makes it harder to have independent clients following independent limits if that’s actually what you want… but I think that’s probably rare. I need to noodle on this a bit more, but I’m guessing the fancier locking class mentioned above might work better for doing both cases, but it’s probably better to tackle that as a separate PR.

Mr0grog:
OK, I think this is probably good. Last thing to do before merging is add release notes.

PR #83, opened 2021-10-13
Fix handling undefined length in CdxRecord

8W9aG:
* Currently there is a bug that prevents the parsing of CdxRecord when length is "-".
* The following code has the following error:

```python
import wayback
client = wayback.WaybackClient()
records = client.search("www.cnn.com/*", matchType="domain", filter_field="statuscode:200")
len(list(records))
```

```
wayback.exceptions.UnexpectedResponseFormat: Could not parse CDX output: "com,cnn)/ 20100218025451 http://www.cnn.com/ text/html 200 JK7O4OWOMDJHBNQJ5U43X5OQRORF6FQB -" (query: {'url': 'www.cnn.com/*', 'matchType': 'domain', 'filter': 'statuscode:200', 'showResumeKey': 'true', 'resolveRevisits': 'true'})
```

The exception occurs on line 542 of `_client.py`. The problem is the service returns "-" where length should be, probably indicating that for some reason this data is missing, however the wayback client attempts to parse it to an int. In this case, use None instead.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -65,6 +65,7 @@ Thanks to the following people for their contributions and help on this package!
 
 - `Dan Allan <https://github.com/danielballan>`_ (Code, Tests, Documentation, Reviews)
 - `Rob Brackett <https://github.com/Mr0grog>`_ (Code, Tests, Documentation, Reviews)
+- `Will Sackfield <https://github.com/8W9aG>`_ (Code, Tests)
 - `Ed Summers <https://github.com/edsu>`_ (Code, Tests)
 - `Lion Szlagowski <https://github.com/LionSzl>`_ (Code, Tests)
 


CONTRIBUTING.rst
```diff
@@ -79,10 +79,10 @@ Ready to contribute? Here's how to set up `wayback` for local development.
 5. When you're done making changes, check that your changes pass flake8 and the tests, including testing other Python versions with tox::
 
     $ flake8 wayback tests
-    $ python setup.py test
+    $ pytest
```
Mr0grog, line 82: Oh, thanks for catching this!

wayback/tests/test_client.py
```diff
@@ -190,6 +190,30 @@ def test_search_removes_malformed_entries(requests_mock):
...
+        requests_mock.get('http://web.archive.org/cdx/search/cdx'
+                          '?url=www.cnn.com%2F%2A'
+                          '&matchType=domain&filter=statuscode%3A200'
+                          '&showResumeKey=true&resolveRevisits=true',
+                          [{'status_code': 200, 'text': bad_cdx_data}])
+        records = client.search('www.cnn.com/*',
+                                matchType="domain",
+                                filter_field="statuscode:200")
+
+        assert 5 == len(list(records))
```
Mr0grog, line 214: This should probably also be asserting that the `.length` property on the last one is `None` while the others are ints, e.g:

```suggestion
        items = list(records)
        assert 5 == len(items)
        assert isinstance(items[0].length, int)
        assert items[4] is None
```

wayback/_client.py
```diff
@@ -539,7 +539,7 @@ def search(self, url, *, matchType=None, limit=None, offset=None,
                     status_code = None
                 else:
                     status_code = int(data.status_code)
-                length = int(data.length)
+                length = None if data.length == "-" else int(data.length)
```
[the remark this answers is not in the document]
8W9aG: Will do

PR #85, opened 2021-10-18
Fix bad timestamps

8W9aG:
* Currently there is a bug that prevents the parsing of CdxRecord when the timestamp is invalid, for example "20000800241623". This timestamp doesn't really make sense because the day is 00 and the hour is 24.
* The following code has the following error:
```python
import wayback
client = wayback.WaybackClient()
records = client.search("www.usatoday.com/*", matchType="domain", filter_field="statuscode:200")
len(list(records))
```
```python
wayback.exceptions.UnexpectedResponseFormat: Could not parse CDX output: "com,usatoday)/2000/century/tech/004.htm 20000800241623 http://www.usatoday.com:80/2000/century/tech/004.htm text/html 200 PAJWSPCRQMVBTYWV4NPJPNDQHKWJC3OO 6177" (query: {'url': 'www.usatoday.com/*', 'matchType': 'domain', 'filter': 'statuscode:200', 'showResumeKey': 'true', 'resolveRevisits': 'true'})
```
The exception occurs on line 549 of _client.py. The problem is the service returns an invalid timestamp, for reasons I cannot fathom, however this tends to happen only on older logs probably indicating it is an issue with certain records, however the wayback client attempts to parse it to an datetime with no correction attempt. In this case, attempt to correct the flaws before parsing.

### DIFF
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -540,15 +540,7 @@ class WaybackClient(_utils.DepthCountedContext):
                 else:
                     status_code = int(data.status_code)
                 length = None if data.length == '-' else int(data.length)
-                # Fix bad timestamps
-                timestamp_chars = list(data.timestamp)
-                # If the timestamp has a day of "00"
-                if timestamp_chars[6:8] == ['0', '0']:
-                    timestamp_chars[7] = '1'
-                # If the timestamp has an hour of "24"
-                if timestamp_chars[8:10] == ['2', '4']:
-                    timestamp_chars[9] = '3'
-                capture_time = _utils.parse_timestamp("".join(timestamp_chars))
+                capture_time = _utils.parse_timestamp(data.timestamp)
             except Exception as err:
                 if 'RobotAccessControlException' in text:
                     raise BlockedByRobotsError(query["url"])
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -540,7 +540,15 @@ class WaybackClient(_utils.DepthCountedContext):
                 else:
                     status_code = int(data.status_code)
                 length = None if data.length == '-' else int(data.length)
-                capture_time = _utils.parse_timestamp(data.timestamp)
+                # Fix bad timestamps
+                timestamp_chars = list(data.timestamp)
+                # If the timestamp has a day of "00"
+                if timestamp_chars[6:8] == ['0', '0']:
+                    timestamp_chars[7] = '1'
+                # If the timestamp has an hour of "24"
+                if timestamp_chars[8:10] == ['2', '4']:
+                    timestamp_chars[9] = '3'
+                capture_time = _utils.parse_timestamp("".join(timestamp_chars))
             except Exception as err:
                 if 'RobotAccessControlException' in text:
                     raise BlockedByRobotsError(query["url"])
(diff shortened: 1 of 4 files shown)


wayback/_client.py
```diff
@@ -540,7 +540,15 @@ def search(self, url, *, matchType=None, limit=None, offset=None,
...
-                capture_time = _utils.parse_timestamp(data.timestamp)
+                # Fix bad timestamps
+                timestamp_chars = list(data.timestamp)
+                # If the timestamp has a day of "00"
+                if timestamp_chars[6:8] == ['0', '0']:
+                    timestamp_chars[7] = '1'
+                # If the timestamp has an hour of "24"
+                if timestamp_chars[8:10] == ['2', '4']:
+                    timestamp_chars[9] = '3'
+                capture_time = _utils.parse_timestamp("".join(timestamp_chars))
```
Mr0grog, line 551: This is complex enough that it’s probably better to put it in `_utils.parse_timestamp()` or in a helper function (e.g. `correct_invalid_timestamp()` or something).

Side note: single quotes on the `join()` call, please! :)

Mr0grog: Also, I *think* the right hour for this is probably `0`, not `23`. It reads to me like a rollover issue, where hour 24 should be the 0th hour of the next day. Does that seem right to you?

Mr0grog (changes requested):
Also, I thought I’d already hit a lot of the weird cases, but you’re finding issues I haven’t seen left and right! Nice find. I am clearly working on too narrow a dataset. 😜

I posted on the Internet Archive Slack about this to see if they had any guidance on the most correct way to handle this case, since they might know of more ways the timestamp could be malformed we should account for, or if this just means the record is bad and we should skip it (for example, no Memento is returned for this URL + timestamp). I’ll probably wait a little bit to see if I get some feedback from them before merging. That said, I think what you’re doing here makes sense.

wayback/_client.py
Mr0grog: Welp, there’s no clear answer on the right general solution here yet, but the IA folks dug under the hood and it seems like the right correction here is actually to *drop* the `00` part from the timestamp in this case, so we should be trying to parse `200008241623` (the underlying record’s raw timestamp is `2000080024162315` [which has too many digits] and the server-sent `Date` header is `Thu, 24 Aug 2000 16:21:19 GMT`, so it looks like some software error caused an extra `00` to be shoved in the middle). More info to come here as I receive it.

8W9aG: There are 2 ways I could solve this, attempt to parse the string and upon failure apply the changes, then parse again in the utils function. Another way would be to detect the amount of digits and pre-emptively correct it, then parse it. Which would you prefer?

Mr0grog: > detect the amount of digits

Oh, sorry, I guess I was not clear. We can’t do this. The raw timestamp with too many digits isn’t available to us — we only know that was the case here because someone at the Internet Archive looked at the raw file that the CDX record was generated from. That raw data is not publicly available (or at least not easily).

> attempt to parse the string and upon failure apply the changes, then parse again in the utils function

I think what you’re doing right now to detect and solve the situation (looking for a `00` in the day-of-month part of the timestamp) is great! I just meant that you should put all that code into its own function, or put it in `_utils.parse_timestamp()` instead of in `WaybackClient.search()`.

Mr0grog (approved):
Sorry I completely lost track of this. Archive folks did not wind up ever having any better thoughts about general approaches here, so I think this is good!

PR #89, opened 2022-09-24
Correct CDX API timestamps with month of 00

### LINKED ISSUE #88: ValueError: time data
I happened to be doing this:

```python

ia = WaybackClient()
for result in ia.search('lapdonline.org', matchType='prefix'):
    print(result)
```

and noticed that after running for 10 minutes or so it blew up with:

```python
Traceback (most recent call last):
  File "/Users/edsummers/Projects/wayback/wayback/_client.py", line 543, in search
    capture_time = _utils.parse_timestamp(data.timestamp)
  File "/Users/edsummers/Projects/wayback/wayback/_utils.py", line 57, in parse_timestamp
    .strptime(''.join(timestamp_chars), URL_DATE_FORMAT)
  File "/usr/local/Cellar/python@3.10/3.10.6_1/Frameworks/Python.framework/Versi
ons/3.10/lib/python3.10/_strptime.py", line 568, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "/usr/local/Cellar/python@3.10/3.10.6_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/_strptime.py", line 349, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data '20000008241731' does not match format '%Y%m%d%H%M%S'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/edsummers/Projects/wayback/./x.py", line 6, in <module>
    for result in ia.search('lapdonline.org', matchType='prefix'):
  File "/Users/edsummers/Projects/wayback/wayback/_client.py", line 547, in search
    raise UnexpectedResponseFormat(
wayback.exceptions.UnexpectedResponseFormat: Could not parse CDX output: "org,lapdonline)/community/op_valley_bureau/north_hollywood/map/map.htm 20000008241731 http://www.lapdonline.org:80/community/op_valley_bureau/north_hollywood/map/map.htm text/html 200 2GPKQMU3BLZXOEZ5EWDQEYHPMKWEHNT3 1158" (query: {'url': 'lapdonline.org', 'matchType': 'prefix', 'showResumeKey': 'true', 'resolveRevisits': 'true'})
```

It looks like the CDX API returned a datetime `20000008241731` which throws an exception during parse because `00` isn't a valid month?

I don't know what the solution is here:

- ignore the record?
- see if the new CDX API is better behaved and switch to it?
- something else?

edsu:
This commit extends the existing logic for handling invalid days of `00` to months that are `00`. It also adds a warning to be logged in both situations.

So if a timestamp of `20200001120000` will get rewritten to `20200112000000` prior to conversion to a datetime.

I have tested on live CDX API data that was failing, and this fix causes the full result to be returned. If more information is known about why this approach is taken it would be good to add in a comment?

Closes #88

### DIFF
--- a/wayback/_utils.py
+++ b/wayback/_utils.py
@@ -50,25 +50,25 @@ def parse_timestamp(time_string):
     Given a Wayback-style timestamp string, return an equivalent ``datetime``.
     """
     # Before parsing, try to fix invalid timestamps.
-    # We've seen a handful of timestamps where a "00" was inserted before either
-    # either the month or day portion of the timestamp, e.g:
+    # We've seen a handful of timestamps where "00" was inserted before the
+    # month or day part of the timestamp, e.g:
     #
     #   20000008241731
     #       ^^ Month is "00"
     #
-    # The Wayback team has looked into several of these, and the "00" was always
-    # *inserted* erroneously, pushing the month or day and the following
-    # components of the timestamp out by two characters. Then the seconds get
-    # truncated (there's only room for 14 characters in the timestamp in the
-    # CDX index). For example:
+    # The Wayback team looked into some of these, and the "00" was always
+    # an insertion, pushing the month or day and the following components of
+    # the timestamp out by two characters. Then the seconds get truncated
+    # (there's only room for 14 characters in the timestamp in the CDX index).
+    # For example:
     #
     #   In raw data:   2000000824173151 (16 characters)
     #   In CDX index:  20000008241731   (Truncated to 14 characters)
     #   Correct value: 20000824173151   (Aug. 24, 2000 at 17:31:51 UTC)
     #
-    # The best we can do for these cases is pull out the erroneous "00" and
-    # substitute "00" for the seconds that got truncated. This isn't exact, but
-    # we can't see the raw data and this is as close as we can get.
+    # The best we can do for these cases is pull out the incorrect "00" and add
+    # "00" for the seconds that got truncated. This isn't exact, but we can't
+    # see the raw data so this is as close as we can get.
     #
     # The issue seems to be limited to some crawls in the year 2000.
     timestamp_chars = list(time_string)
--- a/wayback/_utils.py
+++ b/wayback/_utils.py
@@ -10,6 +10,7 @@ import time
 import urllib.parse
 from .exceptions import SessionClosedError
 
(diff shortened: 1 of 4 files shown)


wayback/_utils.py
```diff
@@ -1,6 +1,7 @@
 from collections import defaultdict
 from contextlib import contextmanager
 from datetime import date, datetime, timezone
+import logging
```
Mr0grog, line 4: Can you please also create a logger so people know *where* this warning is coming from if they hit it? e.g:

```py
logger = logging.getLogger(__name__)
```

Mr0grog (changes requested):
This looks great, excepting one small tweak. Thanks for adding the warnings; I’m not sure why we didn’t think to do that originally.

> If more information is known about why this approach is taken it would be good to add in a comment?

Yes! Looking back at the original PR this adds to (#85), the existing comments could have been much clearer. Feel free to add to them, or I will before merging.

I’m going to wait on this a little so we can get some feedback from Internet Archive folks, since last time (#85) the right answer turned out not to be anything like what we expected.

Mr0grog:
I'll also have to figure out what’s going on with the docs here. :\

wayback/_utils.py
edsu: Nice catch, logger is configured now.

edsu:
I tried to address the docs build over in #90. I'm not sure if that's the right approach though, so of course please feel free to close if not.

Mr0grog:
Heard back from Wayback folks here: https://internetarchive.slack.com/archives/C4PMRAN00/p1664039239024329

And it seems like our solution is the correct one for now. I should have time tomorrow (Sept. 29) to get this merged in (after we resolve the docs issue) and cut a release.

Mr0grog (approved):
I fixed the docs issue in #91, rebased this, and added a more explanatory comment in the code (as you suggested @edsu, thanks for prompting!).

This ought to be good to go, but I’ll wait to merge until tomorrow, when I have less tired eyes to double-check this with.

Mr0grog:
@edsu FYI, this is released in v0.3.3!

edsu:
Thank you @Mr0grog !

PR #90, opened 2022-09-25
Unpin docs dependencies

edsu:
The sphinx doc build run by CircleCI is currently failing:

```
make: Entering directory '/home/circleci/wayback/docs'
Traceback (most recent call last):
  File "/home/circleci/venv/bin/sphinx-build", line 5, in <module>
    from sphinx.cmd.build import main
  File "/home/circleci/venv/lib/python3.8/site-packages/sphinx/cmd/build.py", line 25, in <module>
    from sphinx.application import Sphinx
  File "/home/circleci/venv/lib/python3.8/site-packages/sphinx/application.py", line 42, in <module>
    from sphinx.registry import SphinxComponentRegistry
  File "/home/circleci/venv/lib/python3.8/site-packages/sphinx/registry.py", line 24, in <module>
    from sphinx.builders import Builder
  File "/home/circleci/venv/lib/python3.8/site-packages/sphinx/builders/__init__.py", line 26, in <module>
    from sphinx.util import import_object, logging, rst, progress_message, status_iterator
  File "/home/circleci/venv/lib/python3.8/site-packages/sphinx/util/rst.py", line 22, in <module>
    from jinja2 import environmentfilter
ImportError: cannot import name 'environmentfilter' from 'jinja2' (/home/circleci/venv/lib/python3.8/site-packages/jinja2/__init__.py)
make: *** [Makefile:20: html] Error 1
make: Leaving directory '/home/circleci/wayback/docs'

Exited with code exit status 2

CircleCI
```

It looks like this might be the result of jinja2 imports being deprecated?

https://github.com/sphinx-doc/sphinx/issues/10291

I tried unpinning all the dependencies, and making a few minor changes to the sphinx config and the docs seem to build fine locally but maybe I'm missing some detail about theme/css?

Unfortunately I did have to tell pip to ignore sphinx 5.2.0.post0 which appears to have a recent bug that hopefully will be fixed shortly:

https://github.com/readthedocs/sphinx_rtd_theme/issues/1343

PR #91, opened 2022-09-30
Repair docs build

Mr0grog:
This takes over for #90 and sets appropriate, working constraints on all the dev dependencies involved in building docs (we should really set constraints on *all* the dev dependencies, but this is at least an improvement). I should get the docs building correctly again, and also takes advantage of new features in the docs tools we use (like a new recommended syntax for defaults in numpydoc!).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,11 +110,11 @@ jobs:
   docs:
     working_directory: ~/wayback
     docker:
-      - image: cimg/python:3.10
+      - image: cimg/python:3.8
     steps:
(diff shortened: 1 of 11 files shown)


Mr0grog:
Sheesh! Between the Sphinx vs. Flake8 compatibility issue, getting the new-ish features we want, and incorrect error messages from Read The Docs, this was a lot messier than I expected! At least we have more tightly constrained versions that *should* be less likely to break. (If we have future issues, it may be time to just pin specific versions.)

PR #94, opened 2022-10-26
Overhaul `search()` signature and docs

### LINKED ISSUE #65: Add a default `limit` to `WaybackClient.search()`
Search pagination does not always function correctly if the `limit` parameter is not set, so it might be a good example to set a [high] default value.

For example, @edsu ran into this with a query like:

```py
# Equivalent to:
# http://web.archive.org/cdx/search/cdx?matchType=prefix&showResumeKey=true&url=twitter.com/realDonaldTrump/status/
client.search('twitter.com/realDonaldTrump/status/', matchType='prefix')
```

There’s no resume key in the response. But there are definitely more records than are returned: if you add `limit=100000` in there, you get a resume key, and can page through results, eventually getting far more than were returned in the query without limit.

```py
# Equivalent to:
# http://web.archive.org/cdx/search/cdx?matchType=prefix&showResumeKey=true&limit=100000&url=twitter.com/realDonaldTrump/status/
client.search('twitter.com/realDonaldTrump/status/', matchType='prefix', limit=100_000)
```

I thought we tested this when originally writing the `search` code, so I’m not sure if something has changed/broken or if I’m just misremembering. Either way, the result is really unintuitive, so it might be a good idea to add a large default for `limit` instead of `None` (I’m thinking `500_000`). The docs could potentially use some clarification here, too.

I’m also trying to get some insight from the Wayback team about whether this behavior is intended or a bug (in which case maybe no change is needed).

Mr0grog:
⚠️ Breaking Changes! ⚠️

The signature for `search()` was unwieldy, complicated, and included a bunch of parameters that did nothing or that might cause broken behavior and should never be used (e.g. `page` and `pageSize`). While adding a default value for `limit` to resolve #65, I went ahead and cleaned up `search()`:

- Adds a default value for `limit` in order to fix a very bad footgun. (See #65)
- Adds a lot more detail to the docs, explains special formatting for `url` and complex considerations for `limit`.
- Removes non-functional or breakage-prone parameters that were only included because they were part of the HTTP API. In some cases, they are things this library does automatically for you and aren’t useful to adjust (e.g. `gzip`), in others you could break things (e.g. `page` and `pageSize`), and some are implementation details that users shouldn’t be bothered with (e.g. `resumeKey`, `previous_result`).
- Removes the ability to specify arbitrary extra keyword parameters to be passed directly to the API (there are so many ways to break things here; I argued for this originally so we didn’t have to maintain as much, but it’s just not good).
- Makes all parameters use snake_case.

Internally, the only real change is that this is now a loop instead of a recursive call. This was required in order to not expose internal details as parameters, but is also probably better for call stack and memory management on large queries.

Fixes #65.

@edsu are these API changes going to be OK for you? I’ll make sure this is released as 0.4.0 rather than 0.3.x. If you have any feedback on the docs changes, too, I’d appreciate it (you can preview the build docs from the “checks” at the bottom of the PR).

I might also  try to include a first pass at #8 before releasing.


Mr0grog:
Note: the size reduction here is mostly because responses are gzipped by default now, which was not the case before. (I *think* what happened is that the CDX server now respects the `Accept-Encoding` header insteading of only using the `gzip` query parameter. We use the former, and skip the latter.)

PR #95, opened 2022-10-27
Add Python 3.11 to the test matrix

Mr0grog:
Python 3.11 has been released! (We probably should have done this while it was in beta, but ¯\\\_(ツ)\_/¯)


Mr0grog:
Oops, looks like Circle does not have convenience images for 3.11 yet. https://github.com/CircleCI-Public/cimg-python/issues/164

PR #101, opened 2022-10-31
Handle inconsistent header case for Mementos

### LINKED ISSUE #98: Some original headers are getting lost
It looks like something has changed about either Requests or the Wayback Machine, and we are no longer including all the original archived headers in a Memento object’s `headers` property. For example:

```py
c = WaybackClient()

memento = c.get_memento('https://robbrackett.com/', datetime='20220315020402')
memento.headers
# {'Content-Type': 'text/html'}
```

But the value of `memento.headers` should really be something like:

```py
{'date': 'Tue, 15 Mar 2022 02:04:02 GMT', 'server': 'Apache', 'upgrade': 'h2,h2c', 'connection': 'Upgrade, Keep-Alive', 'last-modified': 'Mon, 30 Nov 2020 22:51:03 GMT', 'accept-ranges': 'bytes', 'content-length': '13182', 'vary': 'Accept-Encoding', 'keep-alive': 'timeout=15, max=768', 'Content-Type': 'text/html'}
```

(Based on https://web.archive.org/web/20220315020402id_/http://robbrackett.com/)

Mr0grog:
This fixes #98, which was caused by two changes:

1. The Internet Archive now returns most *archived* header names (i.e. those prefixed with `'x-archive-org-'`) in lower-case.
2. In HTTP/2 (now possible since we are using HTTPS as of #97), *all* headers are lower-case/case-insensitive.

Since that means it’s possible to wind up with different header name casing when running the same code on different systems, I had to go ahead and make the `Memento.headers` attribute case-insensitive (we probably should have done this long ago for user friendliness anyway). I've implemented that using code largely taken from Requests, since their implementation is not public so we can't just use it directly (plus we plan to switch of Requests at some point anyway). If we wind up switching to HTTPX, they *do* have a public implementation we can just use instead of this.

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -25,10 +25,23 @@ This release includes a significant overhaul of parameters for :meth:`wayback.Wa
 - Expanded the method documentation to explain things in more depth and link to more external references.
 
 
-Maintenance
-^^^^^^^^^^^
+Features
+^^^^^^^^
+
+:attr:`wayback.Memento.headers` is now case-insensitive. The keys of the ``headers`` dict are returned with their original case when iterating, but lookups are performed case-insensitively. For example::
+
+  list(memento.headers) == ['Content-Type', 'Date']
+  memento.headers['Content-Type'] == memento.headers['content-type']
+
+(:issue:`98`)
+
+
+Fixes & Maintenance
+^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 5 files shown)


Mr0grog (commented):
Looks like docs.scipy.org is currently down, so intersphinx can’t fetch references. We should wait until it’s back before merging, but the docs build failure isn’t an issue in our code.

Side note: we definitely use the python.org references, but I’m not sure we need the scipy/numpy/pandas/matplotlib ones. We should pull them out if we don’t.

PR #103, opened 2022-11-02
Add `search_v2()` method

### LINKED ISSUE #8: Implement CDX search based on newer `timemap` CDX API
From a conversation on the Internet Archive’s Research Slack today:

> **kenji**
> Igor http://spacex.com/robots.txt has  `Disallow: /includes/` and http://web.archive.org/cdx/search still honors robots.txt exclusion (because it’s served by older wayback machine), while playback ignores robots.txt (served by new wayback machine).
> 
> http://web.archive.org/web/timemap/cdx?url=www.spacex.com&matchType=domain&gzip=false&filter=statuscode:200&to=20041229235959 will give you more results, including those under `/include/` path. /web/timemap/cdx is served by new wayback.
>
> I’m sorry for the confusing, inconsistent results - we’re trying to migrate all services to new wayback
> 
> oh btw, a tip: `to=2004` will be interpreted as `20041231235959` (if you’re not excluding day 30 and 31 on purpose :smile:) (edited)
> 
> **Igor**
> kenji Thank you!
> 
> **mr0grog**
> Oh, I did not know about `/web/timemap/cdx` as opposed to just `/cdx/search/cdx`. Should I be using the former instead of the latter?
> 
> **kenji**
> /web/timemap/cdx is better functionality-wise, but it’s slower than /cdx/search. So I’d suggest /cdx/search as long as it works ok for your purpose.
> 
> **mr0grog**
> ah, ok
> Will need to consider which is the right path. Is there anything that documents the functional differences? e.g. the robots.txt issue would be a hard one to discover
> 
> Do you have a rough sense of how _much_ slower `/web/timemap/cdx` is?
>
> **kenji**
> I don’t have good benchmark result (it’s nice to have), but I find /web/timemap/cdx 10-20% slower for `matchType=exact` query. `matchType=domain` can be much slower.

We need to look into whether we should switch to `/web/timemap/cdx`.

Mr0grog:
🚧 **Work in Progress!** 🚧

This adds support for the Internet Archive's new, beta CDX search endpoint at `/web/timemap/cdx`. It deals with pagination much better and is eventually slated to replace the search currently at `/cdx/search/cdx`, but is a little slower and still being tested. Fixes #8.

There are still a bunch of things to be done before merging:

- [x] Get clarity from Internet Archive folks on fields, order (see if they get `output=json` working), etc.
- [ ] Determine whether all the other parameters still work and work the same (`filter`, `collapse`, `resolveRevisits`, etc) and whether there are new ones we can/should use. *(Update: `resolveRevisits` is badly broken, but the rest are the same as original search and work fine. Checking w/ Wayback folks for more detail.)*
- [ ] Implement rate limiting
- [ ] Determine if `search_v2` is the right name, or if it should be something else (`search_beta()`? `search_next()`?)
- [ ] Add tests


wayback/_client.py
```diff
@@ -609,6 +609,250 @@ def search(self, url, *, match_type=None, limit=1000, offset=None,
...
+                else:
+                    query[key] = str(value).lower()
+
+        # Since pages are a number of *blocks searched* and not results, a page
+        # in the middle of the result set may have nothing in it. The only way
+        # to know when to stop iterating is to check how many pages there are.
+        page_count = int(self.session.request('GET', CDX_SEARCH_2_URL, params={
+            **query,
+            'showNumPages': 'true'
+        }).text)
```
Mr0grog, line 758: This might not be necessary! I just discovered that sending too high a `page` value gets a 400 error with the header `x-archive-wayback-runtime-error: page must be smaller than numpages`, so we can in theory check for that and stop.

That said, that’s a very human-readable message and feels unstable. We should check with folks at the Internet Archive about what approach they’d prefer people use.

Mr0grog:
Update: the way you control output format in the new search is not with `?output=<format>` but instead with the path: `/web/timemap/<format>`. So the right way to get JSON is:

```
GET /web/timemap/json?url=<url>&other=params&go=here
```

Mr0grog:
Some updates:

- Most of the other parameters continue to work just as they did. But `resolveRevisits` seems to break searches. I’ve ignored it in the code for now and asked Wayback folks about whether it planned or no longer intended to work, and will figure out the right way to handle it after I hear back.

- I added copies of most of the `search()` tests. I think I should probably re-organize these so they are interleaved with the original search tests (so it’s easier to make sure a new test covers both) or even [parametrize](https://docs.pytest.org/en/7.2.x/how-to/parametrize.html) some of these tests so one test covers both implementations. (Not feasible in cases where we are exercising behavior or arguments that differ between the two.)

- There is a weird issue where the search sometimes returns some very recent results (like just the last week-ish?) even if they are after the `to` timestamp. Our tests currently fail because of this. I’ve asked Wayback folks about it (seems like a bug), and am debating whether the best behavior here is to skip or implement the test differently (and expose users to the bug), or to strip the bad results client-side.

At this point, there’s a little more I can do (rate limiting, plus code cleanup), but we the main blocker by lack of clarity on bugs/intended behavior from Wayback, which we’ll have to wait to hear back on.

Mr0grog:
In the time since this was last touched, the Internet Archive finished their rewrite, and then later *replaced* the old API with the new one. (It gained enough features to be largely compatible and has multiple pagination modes, and a predictable set and ordering of columns.)

That means the work here no longer really relevant, so I’m closing this.

There is still more work to do to better support the new version, though. For tracking that, see #205.

PR #108, opened 2022-11-12
Use the `Link` header to get a memento’s URL

### LINKED ISSUE #99: `Memento.url` property can be wrong if it is SURT-equivalent to the actual URL
If you request a memento URL with a SURT form that is equivalent to the memento’s actual URL, the `url` property of the resulting memento object is incorrect — it reflects the URL you requested, rather than the actual, captured URL.

For example:

```py
c = WaybackClient()

memento = c.get_memento('http://robbrackett.com/', datetime='20220315020402')
memento.url
# 'http://robbrackett.com/'
# But the actual capture was from:
# 'https://robbrackett.com/'

# The `link` header has the right info:
memento._raw_headers['link']
# '<https://robbrackett.com/>; rel="original", ...'
```

The right details are in the `link` header, and we should be parsing that. We’ve had a feature request to do that for a while (#57), but I hadn’t realized that there was a bug like this that we *have* to do it to properly work around.

Mr0grog:
This fixes an issue where the `Memento.url` property could be slightly incorrect, since it was based on the URL you requested the memento from (e.g. `https://web.archive.org/web/20221010000000/<url>`), rather than the actual URL the memento was captured from. The URL the memento is requested from matches records via SURT key rather than the URL.

For example, requesting an archived copy of `http://fws.gov/` might return a capture of `https://www.fws.gov/` instead. The returned Memento object’s `url` property used to be `http://fws.gov/` in this case, but this changes it to be `https://www.fws.gov/`.

Fixes #99.

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -2,6 +2,12 @@
 Release History
 ===============
 
+In Development
+--------------
+
+Fix an issue where the :attr:`Memento.url` attribute might not be slightly off (it could have a different protocol, different upper/lower-casing, etc.). (:issue:`99`)
+
+
 v0.4.0 (2022-11-10)
 -------------------
 
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -7,6 +7,39 @@ In Development
 
 Fix an issue where the :attr:`Memento.url` attribute might not be slightly off (it could have a different protocol, different upper/lower-casing, etc.). (:issue:`99`)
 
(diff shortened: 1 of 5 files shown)


Mr0grog:
I think I might hold off on merging this for the weekend and consider adding a `links` property to `Memento` that is the parsed links, since I had to make use of them for this anyway.

It also *might* make sense to parse the `Memento-Datetime` header (we can use Python’s built-in [`email.utils.parsedate_to_datetime` function](https://docs.python.org/3/library/email.utils.html#email.utils.parsedate_to_datetime) for that). OTOH, I don’t see any situation where I’d expect the timestamp we are already parsing from the Memento URL to be inaccurate like the capture URL was, so this may not really be meaningful.

PR #113, opened 2023-02-27
Ensure memento's links match memento's mode

### LINKED ISSUE #111: Memento links should be in same mode as Memento
In #108, I added a `link` property to `Memento` objects with parsed data from the `Link` HTTP headers of mementos. However, the links to *other* mementos in that data turn out to always be in `view` mode, regardless of the mode of the memento you requested!

For example:

```py
from wayback import WaybackClient, Mode
client = WaybackClient()

memento = client.get_memento('https://epa.gov/', '20230210003633')

# Memento is in original mode:
memento.mode == Mode.original.value
# But the links are not:
memento.links == {
    'original': {
        'url': 'https://www.epa.gov/',
        'rel': 'original'
    },
    'timemap': {
        'url': 'https://web.archive.org/web/timemap/link/https://www.epa.gov/',
        'rel': 'timemap',
        'type': 'application/link-format'
    },
    'first memento': {
        # This URL is in `view` mode, not `original`!
        'url': 'https://web.archive.org/web/19970418120600/http://www.epa.gov:80/',
        'rel': 'first memento',
        'datetime': 'Fri, 18 Apr 1997 12:06:00 GMT'
  },
  # ...more links cut for brevity...
}
```

The suggested use for these links is to pass them directly to `get_memento()`, but that might get you a memento in a different mode than you expect! It’s a footgun.

Some options here:

1. Drop the `links` attribute on `Memento` for now. Users can parse the `Link` header(s) themselves if they want it, and are responsible for using them appropriately. (In this case, we also need to reopen #57.)

2. Update the `url` field on any link that references a memento to match the mode of the `Memento` object they are attached to.

    Side note: how do we identify which things are mementos? Look for `"memento"` as a substring in the `rel` field? Look for `url` fields that match known memento URL patterns?

3. Instead of the values in `links` being dictionaries, make them some more useful data object. References to other mementos might be more like our `CdxRecord` objects, where the `url` is the *captured* URL (e.g. `http://www.epa.gov/` instead of the memento URL), the timestamp is a `datetime` object, etc.
    - This one’s pretty complicated! It’s how I envisioned this feature might evolve, but isn’t obviously worthwhile in the short term.
    - I don’t know the complete universe of possible object types (it’s not just mementos, see the first two entries in the example above) and technically what goes here is pretty arbitrary. How do we future-proof things we haven’t modeled yet?

I think (3) has too many open questions, but we should do (1) or (2) before cutting a 0.4.1 release.

Mr0grog:
The `links` attribute on a Memento object links to other related resources, such as the first/previous/next/last memento of the current memento's URL. These are super useful for iterating and navigating through mementos, BUT the archive.org servers always return these as links to the memento in "view" mode. Users of this library rarely use "view" mode, and if you are iterating through links in a different mode, simply following one of the links can lead to subtle mistakes! To mitigate this, we now change all the links that reference mementos to use the same mode as the `Memento` object they are attached to.

Fixes #111.

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -22,22 +22,22 @@ Features
           'rel': 'original'
       },
       'first memento': {
-          'url': 'https://web.archive.org/web/20050323155300/http://www.fws.gov:80/birds',
+          'url': 'https://web.archive.org/web/20050323155300id_/http://www.fws.gov:80/birds',
           'rel': 'first memento',
           'datetime': 'Wed, 23 Mar 2005 15:53:00 GMT'
       },
       'prev memento': {
-          'url': 'https://web.archive.org/web/20210125125216/https://www.fws.gov/birds/',
+          'url': 'https://web.archive.org/web/20210125125216id_/https://www.fws.gov/birds/',
           'rel': 'prev memento',
           'datetime': 'Mon, 25 Jan 2021 12:52:16 GMT'
       },
       'next memento': {
-          'url': 'https://web.archive.org/web/20210321180831/https://www.fws.gov/birds',
+          'url': 'https://web.archive.org/web/20210321180831id_/https://www.fws.gov/birds',
           'rel': 'next memento',
           'datetime': 'Sun, 21 Mar 2021 18:08:31 GMT'
       },
       'last memento': {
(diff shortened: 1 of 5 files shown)


Mr0grog:
I wound up buried in some other work issues last week, circling back around to this now. Planning to release v0.4.1 later this evening.

PR #120, opened 2023-07-11
multiple_filters_client.py

BilibalaX:


wayback/_client.py
```diff
@@ -651,6 +651,8 @@ def search(self, url, *, match_type=None, limit=1000, offset=None,
             if value is not None:
                 if isinstance(value, str):
                     query[key] = value
+                elif isinstance(value, list) and key == 'filter':      # Handle 'filter' field as list
+                    query[key] = value                                 # Assign the list to the 'filter' key
```
Mr0grog, line 655: The code here is pretty self-explanatory, could you please remove the comments?

Mr0grog (changes requested):
Thanks for doing this!

Would you please update the docs on line 573-575?

Also, please add a test for setting `filter_field` to a list and one for setting it to a string.

Mr0grog:
@BilibalaX are you planning to continue working on this, or should I close it out?

PR #122, opened 2023-09-18
Upgrade Versioneer to v0.29

Mr0grog:
The version of Versioneer we were previously on (v0.18) does not appear to be compatible with Python 3.12 (testing with 3.12.0.rc2). Upgrading Versioneer fixes the issue. This sticks with our current install method (vendored) and does not upgrade to Versioneer’s newer “build-time” mode (which is still marked as experimental).


Mr0grog:
Part of #123.

Mr0grog:
Versioneer no longer lists support for Python 3.6 (which we still support here), but it still seems to work fine. And should be a non-issue in release builds, where most of the Versioneer code gets replaced with literal content.

Issue #123, opened 2023-09-18
Ensure support for Python 3.12


Mr0grog:
This package mostly works for Python 3.12 (testing with 3.12.0rc2, which [Python Core expects to be the final release candidate](https://pythoninsider.blogspot.com/2023/09/python-3120-release-candidate-2-now.html)), but has some tooling issues:
- [x] Versioneer does not work (#122)
- [x] Flake8 does not work (they added support in v6.1.0 which is a major release ahead of where we are) (#124)
- [x] Other?
    - [x] Setuptools no longer bundled with Python (#125)

I aim to get all this cleared up this week and cut a release with support on Monday, September 25, 2023, since the planned final release date for Python 3.12.0 is two weeks from today (Monday, October 2, 2023).

Mr0grog:
Published 0.4.3a1 to PyPI, which seems to install and work well on Python 3.12.0rc2 for me. I think we can consider this good.

I’d like to take some time this weekend and try to solve #116 and #119 as well in the final release of v0.4.3 on Monday.

PR #124, opened 2023-09-18
Upgrade Flake8 to v6.1.0

Mr0grog:
Earlier versions of flake8 do not support Python 3.12. Part of #123.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,10 @@ commands:
     parameters:
       python-version:
         type: string
+      install-dev:
+        description: Install dev dependencies (not just test dependencies)
+        type: boolean
+        default: false
       install-docs:
         description: Install docs dependencies
         type: boolean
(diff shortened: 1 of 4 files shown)


Mr0grog:
Well, this version of flake8 does not support Python < 3.8. Those versions are all end-of-life at this point (https://devguide.python.org/versions/) so this is probably OK. Will need to update the checks and classifiers in `setup.py` here, too: https://github.com/edgi-govdata-archiving/wayback/blob/1ea12233cc1586c0af4fe8c311e286a0bb0a2ec1/setup.py#L7-L23

Mr0grog:
Also worth noting: Google Colab has since updated to Python 3.10, so this shouldn't break any Colab stuff.

Mr0grog:
Trying an alternative by splitting out test dependencies from other dev dependencies, kind of like how we handle docs. Maybe more complex than is healthy, but lets us continue being Python 3.6 compatible.

Issue #137, opened 2023-11-01
search_calls_per_second needs to be dialed down


edsu:
I was running some fairly simple data retrieval in [this Notebook](https://github.com/edsu/foiaonline/blob/main/Notebook.ipynb)(see the Wayback section) and I discovered that I got completely blocked from accessing web.archive.org! Luckily I remembered that there was Internet Archive's #wayback-researchers Slack channel, where I got this reponse.

> Hi edsu - I found your /cdx requests from 4:29UTC. Those requests are limited to an average of 60/min. Over that and we start sending 429s. If 429s are ignored for more than a minute we block the IP at the firewall (no connection) for 1 hour, which is what happened to you.  Subsequent 429s over a given period will double that time each occurrence. If you can keeping your api request < 60/minute you will prevent this from happening.

I thought that the openwayback module's defaults would have prevented me from going over the 60 requests per minute (one per second) and I thought wayback's support for handling 429 responses would have backed off sufficiently fast. I suspect that the goal posts on the server side have changed recently because I had code that worked a month or so ago, which stopped working (resulting in the block).

I was able to get around this by using a custom WaybackSession where I set the `search_calls_per_second` to `0.5`, but I suspect `1.0` would probably work better. Maybe the default could me moved down from 1.5 to 1.0?

Also, perhaps there needs to be some logic to make sure to wait a minute when encountering a 429 as well?

Mr0grog:
Keeping this open to track other adjustments mentioned here to handle things better in general here (e.g. when we hit a 429) rather than just adjusting the limits (done in #140).

This is also tied in with a long-running goal I’ve had to revamp rate limiting, since part of the problem here is that we implement the throttle on calls to `search()` and `get_memento()` instead of in the actual `send()` method (or around calls to it), where it *should* be happening. (Elsewhere in EDGI code, we have [a nice `RateLimit` object](https://github.com/edgi-govdata-archiving/web-monitoring-processing/blob/6a2a11b4c93be7fac7e9f23183c6d3bbb0ce1e7a/web_monitoring/utils.py#L225-L274) that works like a lock that I’d like to use here, so we can pass a rate limit into `send()` or otherwise share a single lock around the codebase as needed.)

Mr0grog:
Sketching out some more detailed thoughts here:

[...]

PR #139, opened 2023-11-01
More debug logging

### LINKED ISSUE #138: More logging
I think it would be useful to see more debug logging when requests are being made and when sleeping happens as the result of errors. When I was trying to diagnose #137 I had to sprinkle some `logger.warn` statements in.

edsu:
To help diagnose when things are going wrong this commit adds some additional debug logging when requests are happening, what parameters are being used, sleep behavior and error responses from web.archive.org.

Closes #138

### DIFF
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -408,14 +408,15 @@ class WaybackSession(_utils.DisableAfterCloseSession, requests.Session):
                         raise RateLimitError(result)
                     return result
             except WaybackSession.handleable_errors as error:
-                logger.warn("caught exception during request: %s", error)
                 response = getattr(error, 'response', None)
                 if response:
                     read_and_close(response)
 
                 if retries >= maximum:
                     raise WaybackRetryError(retries, total_time, error) from error
-                elif not self.should_retry_error(error):
+                elif self.should_retry_error(error):
+                    logger.warn("Caught exception during request, will retry: %s", error)
+                else:
                     raise
 
             # The first retry has no delay.
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -404,9 +404,12 @@ class WaybackSession(_utils.DisableAfterCloseSession, requests.Session):
                 result = super().send(*args, **kwargs)
                 if retries >= maximum or not self.should_retry(result):
                     if result.status_code == 429:
-                        logger.warning("caught rate limit error: %s", result.content)
                         raise RateLimitError(result)
                     return result
+                else:
+                    # TODO: parse and use Retry-After header if present.
+                    # TODO: add additional delay for 429 responses.
+                    logger.debug('Received error response (status: %s), will retry', result.status_code)
             except WaybackSession.handleable_errors as error:
                 response = getattr(error, 'response', None)
                 if response:
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -423,7 +423,7 @@ class WaybackSession(_utils.DisableAfterCloseSession, requests.Session):
             if retries > 0:
                 seconds = self.backoff * 2 ** (retries - 1)
                 total_time += seconds
-                logger.debug("retrying after sleep of %s seconds", seconds)
+                logger.debug("Will retry after sleeping for %s seconds...", seconds)
                 time.sleep(seconds)
 
             retries += 1
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -400,7 +400,7 @@ class WaybackSession(_utils.DisableAfterCloseSession, requests.Session):
         retries = 0
         while True:
             try:
-                logger.debug("sending %s", kwargs)
+                logger.debug("sending HTTP request %s '%s', %s", args[0].method, args[0].url, kwargs)
                 result = super().send(*args, **kwargs)
                 if retries >= maximum or not self.should_retry(result):
                     if result.status_code == 429:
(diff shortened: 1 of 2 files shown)


wayback/_client.py
```diff
@@ -712,6 +717,7 @@ def search(self, url, *, match_type=None, limit=1000, offset=None,
             sent_query, next_query = next_query, None
             with _utils.rate_limited(self.session.search_calls_per_second,
                                      group='search'):
+                logger.debug("querying %s %s", CDX_SEARCH_URL, sent_query)
```
Mr0grog, line 720: Hmmm, do you think this is useful when the first thing `WaybackSession.send()` does is log the request?

wayback/_client.py
```diff
@@ -400,17 +400,21 @@ def send(self, *args, **kwargs):
         retries = 0
         while True:
             try:
+                logger.debug("sending %s", kwargs)
                 result = super().send(*args, **kwargs)
                 if retries >= maximum or not self.should_retry(result):
                     if result.status_code == 429:
+                        logger.warning("caught rate limit error: %s", result.content)
```
Mr0grog, line 407: I don’t think we really need to log here, or at least this shouldn’t be a `warning`. If the user (of this library) gets an exception (which they will because we `raise` here), they don’t need a warning logged as well.

wayback/_client.py
```diff
@@ -400,17 +400,21 @@ def send(self, *args, **kwargs):
...
                         raise RateLimitError(result)
                     return result
             except WaybackSession.handleable_errors as error:
+                logger.warn("caught exception during request: %s", error)
                 response = getattr(error, 'response', None)
                 if response:
                     read_and_close(response)
 
                 if retries >= maximum:
+                    logger.error("Too many retries %s >= %s", retries, maximum)
```
Mr0grog, line 417: Same as above here. This raises all the way up to user code, so I do not think logging, or at least not logging at a level above `info` at most, makes a lot of sense at this point.

wayback/_client.py
```diff
@@ -419,6 +423,7 @@ def send(self, *args, **kwargs):
             if retries > 0:
                 seconds = self.backoff * 2 ** (retries - 1)
                 total_time += seconds
+                logger.debug("retrying after sleep of %s seconds", seconds)
```
Mr0grog, line 426: Minor nit: I think it would be helpful if this message were clearer that it is about to sleep, and hasn’t just woken up from sleeping, e.g:

```suggestion
                logger.debug("Will retry after sleeping for %s seconds...", seconds)
```

Or

```suggestion
                logger.debug("Sleeping for %s seconds before retrying...", seconds)
```

wayback/_utils.py
```diff
@@ -226,7 +226,9 @@ def rate_limited(calls_per_second=2, group='default'):
             minimum_wait = 1.0 / calls_per_second
             current_time = time.time()
             if current_time - last_call < minimum_wait:
-                time.sleep(minimum_wait - (current_time - last_call))
+                secs = minimum_wait - (current_time - last_call)
+                logger.debug("sleeping %s", secs)
```
Mr0grog, line 230: I think it would be helpful if this message specified that the cause is the rate limiter, e.g:

```suggestion
                logger.debug("Hit rate limit, sleeping for %s seconds", secs)
```

Mr0grog (commented):
Some minor nits on the messaging, but this looks good. Thanks for flagging and adding a PR for it!

wayback/_client.py
edsu: Good point, I'll remove it.

wayback/_client.py
edsu: I believe 429 errors are raised and then caught. I think it's useful to know when it's happening?

wayback/_client.py
edsu: Ok, removing.

wayback/_client.py
Mr0grog: Where does it get caught?

Mr0grog: OK, so I tested this in practice just to make sure, and the exception we raise at this location does not get caught.

I think the confusion here is that the exception handler here is only catching networking errors, not non-200 responses, and what you were really looking for is a warning that Wayback returned a 429 and we are going to retry. That needs to happen with an `else` on the `if` before this one (`if retries >= maximum or not self.should_retry(result):`). I’m going to go ahead and add that here.

Also, in practice, it seems like the CDX server is not sending a `Retry-After` header to tell us how long to pause for (so it doesn’t match [what I remember](https://github.com/edgi-govdata-archiving/wayback/issues/137#issuecomment-1790980712), but it could be the memento server responds differently and I was thinking of that 🤷 ), but it *does* seem like we could stand to pause a little longer when we get a 429, and that would be a good change to make here, too. I’ll see about doing that as a separate PR so this stays focused on logging.

edsu: Thanks for investigating further @Mr0grog! Is there anything we should communicate to Internet Archive?

Mr0grog: If you’re still talking with them about it, maybe let them know it would be nice if they sent a `Retry-After` header with their 429 responses, but otherwise, no. This is really all just issues on our side.

Mr0grog: Anyway, I’ll try and get this merged and released later today.

PR #140, opened 2023-11-01
Lower search calls per second to 1

### LINKED ISSUE #137: search_calls_per_second needs to be dialed down

edsu:
Fixes #137

### DIFF
--- a/wayback/_client.py
+++ b/wayback/_client.py
@@ -347,7 +347,7 @@ class WaybackSession(_utils.DisableAfterCloseSession, requests.Session):
     user_agent : str, optional
         A custom user-agent string to use in all requests. Defaults to:
         `wayback/{version} (+https://github.com/edgi-govdata-archiving/wayback)`
-    search_calls_per_second : int or float, default: 1.5
+    search_calls_per_second : int or float, default: 1
         The maximum number of calls made to the search API per second.
         To disable the rate limit, set this to 0.
     memento_calls_per_second : int or float, default: 30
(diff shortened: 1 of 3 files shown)


Mr0grog:
>  I suspect that the goal posts on the server side have changed recently because I had code that worked a month or so ago, which stopped working (resulting in the block).

Yeah, this is also more limited than I recall it being when we first implemented, too. Thanks for looking into it!

PR #144, opened 2023-12-01
Try replacing `setup.*` with `pyproject.toml`

Mr0grog:
Direct usage of `setup.py` (that is calling `python setup.py some_command` to do builds and such) has been deprecated and no longer supported for quote a while, and I’ve been meaning to try to move everything to `pyproject.toml` (the new format for declaring package metadata). It *seems* like `pyproject.toml` supports all the nice things now, so this is my experimental attempt to switch to it and drop `setup.py` and `setup.cfg` entirely.

(This seems like a good time to do this, since there are some other semi-breaking changes on tap, like [rethinking how rate limit errors are handled](https://github.com/edgi-govdata-archiving/wayback/issues/137#issuecomment-1828873225).)

This doesn’t quite work yet — the main sticking point is Versioneer. When `pip` or `build` try and make an isolated environment, it’s missing the runtime requirements for the package and can’t import the version from the code.

This _might_ be fixable by:
- Making a separate `_build_version.py` file that pulls in the version info the same as `__init__.py` does but without requiring any other dependencies.
- Bringing back a minimal version of `setup.py` just to run Versioneer. (But it seems like this *shouldn’t* be necessary since Versioneer 0.29 claims support for `pyproject.toml`-only build systems: https://github.com/python-versioneer/python-versioneer/releases/tag/0.29)
- Not doing isolated builds? Same as today, I guess, but disappointing.

Need to experiment here.

Obviously this will need CI changes and maybe Dependabot config, too.

Also Needs testing to make sure builds based on this are still installable and workable on Python 3.6 (assuming a recent version of pip is being used to install).


danielballan:
I was just thinking of dipping a toe back in here, starting with this. May I toss some commits here on the versioneer aspect?

Mr0grog:
Dan!! Absolutely, please do add some commits

Mr0grog:
I was just reading through the PRs that went into v0.29 of Versioneer and https://github.com/python-versioneer/python-versioneer/pull/371 makes it sound like maybe `pyproject.toml`-only is only supported for the pdm backend, and not setuptools, but 🤷

Mr0grog:
@danielballan that sounds pretty reasonable to me! (I think last time I was playing with newer Python tools, I liked PDM more than Hatch, but that was mainly because I didn’t think Hatch managed venvs, and it looks like that’s either changed or I understood it wrong back then. I also liked that it did PEP 582, but that got killed and I am sad.)

One question about Python versions, though:

We currently support Python 3.6, and I think Hatch/Hatchling are 3.7+ (I guess the latest Hatchling looks like just 3.8+). I think the `build-system` deps are only need during builds, so this isn’t a big deal (since I have been building from a newer version), but will this be a problem if a Python 3.6 user winds up installing from an sdist instead of a wheel?

IIRC we support 3.6 here because Google Colab took a long time to update, but I just checked and it looks like Colab is on Python 3.10 now. So it’s probably OK to change our minimum Python version. Just want to flag it as a breaking change to avoid *if it’s not too hard*.

Mr0grog:
> Would that policy create a problem for known users?

Nah, I think that’s ok here. The only holdup for known users (mainly @edsu and @ericnost) was Google Colab, and that’s no longer an issue.

danielballan:
In the ~5 years it's been since I showed my face here, my commit bit went away (which is very reasonable). I started a PR into this branch over on my fork: https://github.com/edgi-govdata-archiving/wayback/pull/145

Mr0grog:
Do you want it back? I’m happy to add you as a committer/maintainer/admin (depending on how responsible you want to be 😉).

danielballan:
Thanks. Let’s give it a couple months and see if I manage to stay active. 🙂

pyproject.toml
```diff
@@ -0,0 +1,93 @@
+[project]
+name = "wayback"
+# version = "0.0.1"
+dynamic=["version"]
+description = "Python API to Internet Archive Wayback Machine"
+# FIXME: validate this is the correct replacement for long_description and
+# long_description_content_type!
+# It seems this can also be:
+# readme = {content-type = "text/x-rst", file = "README.rst"}
+readme = "README.rst"
```
Mr0grog, line 10: Yep, this appears to work: https://test.pypi.org/project/wayback/0.5.0a2/

pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+# TODO: put contributors in `authors` field:
+# authors = [
+#   { name="XYZ" },
+# ]
+# FIXME: how does this show up in PyPI? In setup.py we had:
+#   license="BSD (3-clause)"
+# But that's not the right format here. And there's no trove classifier for it.
+# Maybe this should be:
+#   license = {text = "BSD (3-clause)"}
+license = {file = "LICENSE"}
```
Mr0grog, line 24: Yeah, this should probably change. Here’s how the the file reference shows up on test.pypi.org:

Screenshot from PyPI showing several lines of the full license text

vs. what we currently have with `setup.py`:

Screenshot from PyPI showing just the name of the license

Mr0grog: Changed this and uploaded to test.pypi.org; it fixes the issue!

Mr0grog:
Yowza, this dropped 3,025 lines of code! Granted that was mostly Versioneer, but nice to see it go. 👋

.git_archival.txt
```diff
@@ -0,0 +1,4 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
+ref-names: $Format:%D$
```
webknjaz, line 4: @Mr0grog FYI it was later determined that having `ref-names` is problematic, and it's no longer recommended to include it into this template: https://setuptools-scm.rtfd.io/en/latest/usage/#git-archives / https://github.com/pypa/setuptools_scm/pull/1033.

Mr0grog: Ah, that is interesting, thanks very much. I’ll pull it out of the main development branch. 🙇 

That said, please post things like this as a new issue (or an existing open issue if applicable) rather than re-opening discussion on a completed PR in the future. It makes it a little easier to track things that need to get done (and attribution for release notes). 🙂

webknjaz: Yeah, it was more of a drive-by discovery, a thing that I noticed in passing so I put a comment where some context exist. I'm not a user of this project, but when I notice something in places I visit while doing git paleontology, and don't have time to spend on writing a long explanation, this seems like an acceptable middle ground :)

PR #146, opened 2023-12-04
Stop retrying after rate limit errors

Mr0grog:
We used to retry requests that got a response with a 429 status code (indicating that you've hit a rate limit on the server) automatically, but we've resolved that this is no longer a good approach.

See the discussion in https://github.com/edgi-govdata-archiving/wayback/issues/137#issuecomment-1828873225 for more detail, but the short version is that our previous retry behavior was specifically geared to a workflow EDGI used, and which had custom rate limits and other server behavior that most users won't have, and also helped to work around a bunch of deficiencies in our client-side rate limiting that were solved in the last release (but they do need [a *bit* more work](https://github.com/edgi-govdata-archiving/wayback/issues/137#issuecomment-1790980712) in this release, too). It turns out this led to some users getting blocked, and it's better to leave handling this situation to user code.

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -8,7 +8,9 @@ In Development
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
-- N/A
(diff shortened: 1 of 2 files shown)


Mr0grog (commented):
I’m holding off for now on re-arranging the actual logic (I just took `429` out of the list of retryable statuses) this I’m mulling through whether we should support someone updating the list of retryable statuses at runtime or subclassing `WaybackSession` in order to get retries for rate limit errors.

Issue #58, opened 2020-10-21
Pool connections across threads and/or make WaybackClient thread-safe


Mr0grog:
This is a current major goal, but we didn’t have an issue tracking it!

This package should be thread-safe, but because it is based on [Requests](https://requests.readthedocs.io/), we can’t guarantee that. Requests’s authors won’t guarantee it, and at various times have expressed everything from certainty that it’s *not* thread-safe to cautious optimism that it *might* be, but overall have expressed that they don’t plan to go out of their way to make certain (and then to maintain that status).

Thread safety is important for EDGI’s use: we need to pull *lots* of data and so need to make `get_memento()` calls concurrent. Connections also need to be pooled and shared across threads for stability (when EDGI implemented some hacks to do this, speed and reliability both improved considerably). In any case, other EDGI codebases implement [some pretty nutty workarounds][edgi-hack] to make that *relatively* safe. That shouldn’t be necessary, and **people should just be able to use a single `WaybackClient` across multiple threads.**

**Ideally,** we should switch to a different HTTP client instead of *Requests*. Two options:

1. [urllib3](https://urllib3.readthedocs.io/) is thread-safe and is what Requests is based on, so it should be a reliable switch. Its API is very different, though, so there’s lots to update and test.

2. [httpx](https://python-httpx.org/) is new, has an API that matches *Requests*, and has lots of fancy features. It also has async support in case we ever want to make this package async-capable in the future. However, it’s still in beta (should be final before the end of the year), and may possibly have some funny under-the-hood differences we’ll need to account for. We’d at least need to figure out how to re-implement our [crazy hack for Wayback’s gzip issues](https://github.com/edgi-govdata-archiving/wayback/blob/ee409970c19fcf673cd6c213d188dbca4badff5c/wayback/_client.py#L157-L186).

---

Some other approaches worth keeping in the back pocket, but that probably aren’t ideal:

- A sketched out implementation in #23 makes a funky abstraction that lets it *appear* as though you are using a single `WaybackClient` on multiple threads, when in fact you are using several. It’s clever, but also a little hacky and probably has a lot of messy corner cases as a result. I don’t feel great about it.

[...]

PR #147, opened 2023-12-07
Winter cleaning for rate limiting

### LINKED ISSUE #137: search_calls_per_second needs to be dialed down

Mr0grog:
This significantly refactors how we manage rate limiting with the goal of making it more clear, flexible, and accurate going forward. It encompasses a few major changes:

- [x] **Set limits based on actual values from Internet Archive sysadmins.** We previously determined rate limits via a combination of ad-hoc conversations with Internet Archive staff, our own testing, and feedback from users. I’ve made special effort to reach out to Internet Archive staff and get explicit feedback about both the hard limits and their desired default behavior (80% of hard limits).

- [x] **Replace `rate_limited()` with a `RateLimit` class.** This allows separate sessions to share limits or operate with independent limits in a clearer way. Previously, multiple sessions with different limits would still interact in complicated ways. Limits previously acted more like “how long do I wait after the previous request (from any session, even if it had a different limit).” Now sessions with different limit objects operate completely independently, and sessions that need to share limits can share instances of `RateLimit`.

    The default limits are globally shared `RateLimit` objects, so multiple sessions without explicitly set limits will still coordinate their limiting together just like they used to.

- [x] **Move rate limit application from the client to the session.** We previously applied the rate limits from the client inside the `search()` and `get_memento()` methods. Instead, we’ll apply them inside `WaybackSession.send()`. This ensures that they are always applied (harder to make mistakes) and also fixes an issue where we did no limit following redirects in mementos. The rate limits also belong to the session, so the session is really the right place to apply them anyway.

- [x] Update release notes draft.

Fixes #137.

Some of this also hopefully paves the way for an even bigger refactor of `WaybackSession` in #58.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,10 +20,10 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - cache_v8-wayback-<< parameters.python-version >>-{{ arch }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}-{{ checksum "requirements-docs.txt" }}
-            - cache_v8-wayback-<< parameters.python-version >>-{{ arch }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
-            - cache_v8-wayback-<< parameters.python-version >>-{{ arch }}-{{ checksum "requirements.txt" }}-
-            - cache_v8-wayback-<< parameters.python-version >>-{{ arch }}-
+            - cache_v9-wayback-<< parameters.python-version >>-{{ arch }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}-{{ checksum "requirements-docs.txt" }}
+            - cache_v9-wayback-<< parameters.python-version >>-{{ arch }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
+            - cache_v9-wayback-<< parameters.python-version >>-{{ arch }}-{{ checksum "requirements.txt" }}-
+            - cache_v9-wayback-<< parameters.python-version >>-{{ arch }}-
 
       - run:
           name: Install Dependencies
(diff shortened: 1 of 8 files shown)


Mr0grog:
Most of the work here has been done. Rate limits are now stored internally as a dict indexed by URL path prefix (so they are pretty flexible. I need to decide whether to update the public API to take rate limits set that way as well (see discussion in https://github.com/edgi-govdata-archiving/wayback/issues/137#issuecomment-1845935565; sounds like yes, we should change):

```py
WaybackSession({
    '/web/timemap': _utils.RateLimit.make_limit(timemap_calls_per_second),
    '/cdx': _utils.RateLimit.make_limit(search_calls_per_second),
    '/': _utils.RateLimit.make_limit(memento_calls_per_second),
})
```

PR #155, opened 2023-12-18
Do PyPI release in GHA

danielballan:
Precisely following https://learn.scientific-python.org/development/guides/gha-pure/

This introduces GHA, alongside CircleCI. An additional integration is not ideal, bu there are points in favor:

* GHA, being a GH-internal integration, is not a big jump to make.
* We can use this community-supported recipe as is.
* GHA can handshake directly with PyPI via [Trusted Publisher](https://docs.pypi.org/trusted-publishers/) that we do not need to handle secrets in the environment. (It seems like it ought to be possible to do this with CircleCI, but some cursory Googling did not turn anything up.)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,8 +1,8 @@
 name: CD
 
 on:
-  workflow_dispatch:
-  pull_request:
+  workflow_dispatch: {}
+  pull_request: {}
   push:
     branches:
       - main
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,38 @@
+name: CD
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+  release:
+    types:
+      - published
+
+jobs:
+  dist:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: hynek/build-and-inspect-python-package@v1
+
+  publish:
+    needs: [dist]
+    environment: pypi
+    permissions:
+      id-token: write
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
+
+    steps:
+      - uses: actions/download-artifact@v3
+        with:
+          name: Packages
+          path: dist
+
+      - uses: pypa/gh-action-pypi-publish@release/v1
+
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: hynek/build-and-inspect-python-package@v1
+      - uses: hynek/build-and-inspect-python-package@v2
 
   publish:
     needs: [dist]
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  dist:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: hynek/build-and-inspect-python-package@v1
```
Mr0grog, line 21: Looks like this action is at v2 now (the usage example in the README just hasn't been updated). https://github.com/hynek/build-and-inspect-python-package/releases/tag/v2.0.0

Looks like this upgrade also requires updating `actions/download-artifact` to v4 in the `publish` job.

```suggestion
      - uses: hynek/build-and-inspect-python-package@v2
```

.github/workflows/cd.yml
```diff
@@ -0,0 +1,38 @@
...
+  publish:
+    needs: [dist]
+    environment: pypi
+    permissions:
+      id-token: write
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
+
+    steps:
+      - uses: actions/download-artifact@v3
```
Mr0grog, line 32: To go with the above suggestion about `hynek/build-and-inspect-python-package@v2`:

```suggestion
      - uses: actions/download-artifact@v4
```

.github/workflows/cd.yml
```diff
@@ -0,0 +1,38 @@
+name: CD
+
+on:
+  workflow_dispatch:
+  pull_request:
```
Mr0grog, line 5: Hmmm, this did not run on the PR, and did not give me the button to approve running it (since you are making the PR from a fork), either… do these maybe need to be:

```suggestion
  workflow_dispatch: {}
  pull_request: {}
```

Mr0grog (commented):
Nice, I keep meaning to try out the trusted publishing stuff, but it seems like a good idea. I’m not super excited to split workflows across CircleCI and GHA, but that seems like a good reason to do so.

Just to make sure I’m clear on the intended usage, the way this is meant to work is:
- Builds on every PR update, every push to main, or when run manually
- To publish a release:
    1. Add and push a tag
    2. Manually draft a release on github.com based on the tag
    3. This will automatically build & push the release the PyPI

?

Should we also drop the `build` job from CircleCI, since it seems duplicative? https://github.com/edgi-govdata-archiving/wayback/blob/a513bccd809d14cd587ef6ddf78ca2b279878258/.circleci/config.yml#L104-L123

Mr0grog:
That did not seem to make the workflow run here. It might just need to land in this repo (and not a fork) before it be run, or GH has to see *some* workflows in this repo before it understands that the workflow engine should kick in.

Anyway, I’m cool to ship this if you think it’s good, assuming I got [the above stuff about intended usage](https://github.com/edgi-govdata-archiving/wayback/pull/155#pullrequestreview-1787366729) correct.

danielballan:
I think I've seen this before: GH will not run a _new_ workflow on a PR  branch.

Yes, that summary of the intended usage is precisely right.

I have just pushed a commit to remove the `build` section from the CircleCI config.

Mr0grog (approved):
Looks good, then, merging this in! :shipit:

PR #157, opened 2023-12-30
Remove `requests` from public interfaces

Mr0grog:
`WaybackSession` was originally a subclass of `requests.Session`, and that turns out to have been a poor decision (see https://github.com/edgi-govdata-archiving/wayback/pull/152#issuecomment-1858800671). We wanted to make customizing HTTP behavior relatively straightforward, but using a subclass-able version of `requests.Session` wound up baking problematic parts of requests into our API. In particular, requests is not thread-safe and this setup makes it hard to wrap with thread-safe code. We also unintentionally broke the promise of easy subclassing when we added complex retry and rate limiting behavior.

This is a major, breaking change designed to largely remove requests from our public API:

- `WaybackSession` is now its own, standalone class instead of a subclass of `requests.Session`. It handles stuff like rate limiting, retries, etc.

- `WaybackHttpAdapter` handles making a single HTTP request and is the interface through which we use the requests library.
    - It has a very small API (2 methods: `.request()` sends an HTTP request; `.close()` closes all its network connections).

    - The `.request()` method returns a `WaybackHttpResponse` object, which models just the parts of the response we need. The `WaybackHttpResponse` class is an abstract base class that represents the API that the rest of Wayback needs, and internally we have a `WaybackRequestsResponse` subclass that wraps a response object from Requests in a thread-safe way.

    - Ideally, the small API will make it easy for us to build an alternate implementation with urllib3, httpx, etc, or just to add thread-safe handling of requests. It might also be a better way to let users customize HTTP behavior than the old approach, although you can’t yet supply a custom subclass to `WaybackClient` or `WaybackSession`.

❗ **Open question:** I don’t think that `WaybackSession` should continue to exist. It makes the API complex and it lets the details of requests leak out a little bit through exception handling. Any feedback on this is welcome (cc @danielballan). Some ideas:

1. Move its functionality directly into `WaybackHttpAdapter`.
    - Upside: this gets all the requests stuff — including exception handling — tucked away into one place.
    - Downside: Safely customizing `WaybackHttpAdapter` gets more complex because you have to work around all that machinery. It’s not as straightforward as just overriding the `.request()` method. Some ways to address that:
        - Add a `WaybackHttpAdapter._request()` method (not the underscore) that you can override if you just want to customize the sending of a single HTTP request.
        - Re-architect that functionality into some kind of service/helper object that you can use from a customized `WaybackHttpAdapter` subclass. I think this might be a lot more work than it's worth, though.

2. Move its functionality into `WaybackClient`.
    - Upside: keeps all the things that are special about dealing with HTTP requests at a high level out of the lower-level adapter tooling. Keeps adapters as simple as possible.
    - Downside: Lets some exception handling that is specific to requests leak out into the rest of the package. Makes it harder to change that behavior. What if httpx or some other tool incorporates rate limiting in the future? It would be nice to let an adapter delegate those tasks to the lower-level tool it is acting as glue for.

I think I’m leaning towards some version of (1) here.

This paves the way towards fixing #58, but does not actually solve the issue (`WaybackHttpAdapter` is not yet thread-safe, and needs to use some of the tricks from #23).

### DIFF
--- a/src/wayback/_client.py
+++ b/src/wayback/_client.py
@@ -272,8 +272,8 @@ def clean_memento_links(links, mode):
 
 class WaybackSession(_utils.DisableAfterCloseSession):
     """
-    A custom session object that pools network connections and resources for
-    requests to the Wayback Machine.
+    Sessions manage HTTP requests and network traffic to web servers, adding
+    functionality like retries and rate limiting.
 
     Parameters
     ----------
@@ -463,7 +463,7 @@ class WaybackSession(_utils.DisableAfterCloseSession):
 
 
 # TODO: add retry, backoff, cross_thread_backoff, and rate_limit options that
-# create a custom instance of urllib3.utils.Retry
+# create a custom instance of urllib3.utils.Retry?
 class WaybackClient(_utils.DepthCountedContext):
     """
     A client for retrieving data from the Internet Archive's Wayback Machine.
--- a/src/wayback/_client.py
+++ b/src/wayback/_client.py
@@ -19,9 +19,7 @@ from enum import Enum
 import hashlib
 import logging
 import re
-from requests.exceptions import (ChunkedEncodingError,
-                                 ConnectionError,
-                                 ContentDecodingError,
+from requests.exceptions import (ConnectionError,
                                  ProxyError,
                                  RetryError,
                                  Timeout)
@@ -33,7 +31,7 @@ from urllib3.exceptions import (ConnectTimeoutError,
 from warnings import warn
 from . import _utils, __version__
 from ._models import CdxRecord, Memento
-from ._http import WaybackHttpAdapter
+from ._http import InternalHttpResponse, WaybackHttpAdapter
 from .exceptions import (WaybackException,
                          UnexpectedResponseFormat,
                          BlockedByRobotsError,
@@ -147,7 +145,7 @@ def is_malformed_url(url):
     return False
 
 
-def is_memento_response(response):
+def is_memento_response(response: InternalHttpResponse) -> bool:
     return 'Memento-Datetime' in response.headers
 
 
@@ -155,18 +153,6 @@ def cdx_hash(content):
     if isinstance(content, str):
         content = content.encode()
     return b32encode(hashlib.sha1(content).digest()).decode()
-
-
-def read_and_close(response):
-    # Read content so it gets cached and close the response so
-    # we can release the connection for reuse. See:
-    # https://github.com/psf/requests/blob/eedd67462819f8dbf8c1c32e77f9070606605231/requests/sessions.py#L160-L163
-    try:
-        response.content
-    except (ChunkedEncodingError, ContentDecodingError, RuntimeError):
-        response.raw.read(decode_content=False)
-    finally:
-        response.close()
 
 
 REDIRECT_PAGE_PATTERN = re.compile(r'Got an? HTTP 3\d\d response at crawl time', re.IGNORECASE)
@@ -368,7 +354,7 @@ class WaybackSession(_utils.DisableAfterCloseSession):
         # with Wayback's APIs, but urllib3 logs a warning on every retry:
         # https://github.com/urllib3/urllib3/blob/5b047b645f5f93900d5e2fc31230848c25eb1f5f/src/urllib3/connectionpool.py#L730-L737
 
-    def request(self, method, url, timeout=-1, **kwargs):
+    def request(self, method, url, timeout=-1, **kwargs) -> InternalHttpResponse:
         super().request()
         start_time = time.time()
         timeout = self.timeout if timeout is -1 else timeout
@@ -393,16 +379,16 @@ class WaybackSession(_utils.DisableAfterCloseSession):
 
                 if retries >= maximum or not self.should_retry(response):
                     if response.status_code == 429:
-                        read_and_close(response)
+                        response.close()
                         raise RateLimitError(response, retry_delay)
                     return response
                 else:
                     logger.debug('Received error response (status: %s), will retry', response.status_code)
-                    read_and_close(response)
+                    response.close(cache=False)
             except WaybackSession.handleable_errors as error:
                 response = getattr(error, 'response', None)
                 if response is not None:
-                    read_and_close(response)
+                    response.close()
 
                 if retries >= maximum:
                     raise WaybackRetryError(retries, time.time() - start_time, error) from error
@@ -477,6 +463,8 @@ class WaybackClient(_utils.DepthCountedContext):
     ----------
     session : WaybackSession, optional
     """
+    session: WaybackSession
+
     def __init__(self, session=None):
         self.session = session or WaybackSession()
 
@@ -705,7 +693,7 @@ class WaybackClient(_utils.DepthCountedContext):
             # worry about it. If we don't raise here, we still want to
             # close the connection so it doesn't leak when we move onto
             # the next of results or when this iterator ends.
-            read_and_close(response)
+            response.close()
             if response.status_code >= 400:
                 if 'AdministrativeAccessControlException' in response.text:
                     raise BlockedSiteError(query['url'])
@@ -930,7 +918,7 @@ class WaybackClient(_utils.DepthCountedContext):
             current_url, current_date, current_mode = _utils.memento_url_data(response.url)
 
             # In view mode, redirects need special handling.
-            redirect_url = response.next and response.next.url
+            redirect_url = response.redirect_url
             if current_mode == Mode.view.value and not redirect_url:
                 redirect_url = detect_view_mode_redirect(response, current_date)
                 if redirect_url:
@@ -1006,7 +994,7 @@ class WaybackClient(_utils.DepthCountedContext):
                             playable = True
 
                 if not playable:
-                    read_and_close(response)
+                    response.close()
                     message = response.headers.get('X-Archive-Wayback-Runtime-Error', '')
                     if (
                         ('AdministrativeAccessControlException' in message) or
@@ -1020,7 +1008,7 @@ class WaybackClient(_utils.DepthCountedContext):
                         raise BlockedByRobotsError(f'{url} is blocked by robots.txt')
                     elif message:
                         raise MementoPlaybackError(f'Memento at {url} could not be played: {message}')
-                    elif response.ok:
+                    elif response.is_success:
                         # TODO: Raise more specific errors for the possible
                         # cases here. We *should* only arrive here when
                         # there's a redirect and:
@@ -1037,7 +1025,7 @@ class WaybackClient(_utils.DepthCountedContext):
 
             if redirect_url:
                 previous_was_memento = is_memento
-                read_and_close(response)
+                response.close()
 
                 # Wayback sometimes has circular memento redirects ¯\_(ツ)_/¯
                 urls.add(response.url)
--- a/src/wayback/_client.py
+++ b/src/wayback/_client.py
@@ -19,7 +19,6 @@ from enum import Enum
 import hashlib
 import logging
 import re
-import requests
 from requests.exceptions import (ChunkedEncodingError,
                                  ConnectionError,
                                  ContentDecodingError,
@@ -34,7 +33,7 @@ from urllib3.exceptions import (ConnectTimeoutError,
 from warnings import warn
 from . import _utils, __version__
 from ._models import CdxRecord, Memento
-from ._http import *  # noqa
+from ._http import WaybackHttpAdapter
 from .exceptions import (WaybackException,
                          UnexpectedResponseFormat,
                          BlockedByRobotsError,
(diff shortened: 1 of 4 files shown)


src/wayback/_client.py
```diff
@@ -433,15 +354,16 @@ def __init__(self, retries=6, backoff=2, timeout=60, user_agent=None,
         # with Wayback's APIs, but urllib3 logs a warning on every retry:
         # https://github.com/urllib3/urllib3/blob/5b047b645f5f93900d5e2fc31230848c25eb1f5f/src/urllib3/connectionpool.py#L730-L737
 
-    # Customize the built-in `send()` with retryability and rate limiting.
-    def send(self, request: requests.PreparedRequest, **kwargs):
+    def request(self, method, url, timeout=-1, **kwargs) -> WaybackHttpResponse:
+        super().request()
```
Mr0grog, line 358: This `super()` call is really just to keep `_utils.DisableAfterCloseSession` working, and I think the reality here is that we should probably just get rid of that utility. It provides some nice guardrails, but they don’t offer a huge amount of value.

src/wayback/_client.py
```diff
@@ -539,18 +444,12 @@ def get_retry_delay(self, retries, response=None):
...
     def reset(self):
         "Reset any network connections the session is using."
-        # Close really just closes all the adapters in `self.adapters`. We
-        # could do that directly, but `self.adapters` is not documented/public,
-        # so might be somewhat risky.
         self.close(disable=False)
-        # Re-build the standard adapters. See:
-        # https://github.com/kennethreitz/requests/blob/v2.22.0/requests/sessions.py#L415-L418
-        self.mount('https://', requests.adapters.HTTPAdapter())
-        self.mount('http://', requests.adapters.HTTPAdapter())
```
Mr0grog, line 549: IIRC, I added this to help with some problems we were having managing connections in BIG jobs at EDGI, and that were later solved with better fixes. It’s kind of a weird method, and its presence is even weirder with the refactoring here. I think I should just drop it entirely.

src/wayback/_http.py
```diff
@@ -0,0 +1,282 @@
...
+            Method to use for the request, e.g. ``'GET'``, ``'POST'``.
+        url : str
+            The URL to reqeust.
+        params : dict, optional
+            For POST/PUT requests, data to be encoded in the response body. For
+            other methods, this should be encoded as the querystring.
+        headers : dict, optional
+            The HTTP headers to send with the request.
+        allow_redirects : bool, default: True
+            Whether to follow redirects before returning a response.
```
Mr0grog, line 259: If we keep rate limiting out of the adapter layer, this argument should be removed, since redirect following has to be handled wherever rate limits are (or higher up).

src/wayback/_http.py
```diff
@@ -0,0 +1,282 @@
...
+        timeout : int or float or tuple of (int or float, int or float), optional
+            How long to wait, in seconds, before timing out. If this is a single
+            number, it will be used as both a connect timeout (how long to wait
+            for the first bit of response data) and a read timeout (how long
+            to wait between bits of response data). If a tuple, the values will
+            be used as the connect and read timeouts, respectively.
```
Mr0grog, line 265: I’m a little worried this is too narrowly designed around how requests does it. It's probably not a huge deal, though, and keeps the timeout interface the same-ish as it is today.

src/wayback/tests/test_client.py
```diff
@@ -694,7 +694,7 @@ def test_request_retries(self, requests_mock):
                                               {'text': 'good', 'status_code': 200}])
         session = WaybackSession(retries=2, backoff=0.1)
         response = session.request('GET', 'http://test.com')
-        assert response.ok
+        assert response.is_success
```
Mr0grog, line 697: This change is a bit arbitrary, but I was inspired by httpx's reasoning for the same change: https://www.python-httpx.org/compatibility/#checking-for-success-and-failure-responses

Considering whether we should also do this in the `Memento` class (maybe just a *deprecation* of `.ok` there, though).

src/wayback/_http.py
```diff
@@ -0,0 +1,282 @@
...
+    attribute of an exception, but should otherwise not be exposed to user
+    code in normal circumstances. It's meant to wrap to provide a standard,
+    thread-safe interface to response objects from whatever underlying HTTP
+    tool is being used (e.g. requests, httpx, etc.).
+    """
+    status_code: int
+    headers: dict
+    encoding: Optional[str] = None
+    url: str
+    links: dict
```
Mr0grog, line 94: Wondering if this class should move to `_models`, especially if we make customizing `WaybackHttpAdapter` a thing. (Since you’ll need to make your own subclass of this to return from an adapter that doesn’t use *Requests*.)

src/wayback/_http.py
Mr0grog: Rate limiting wound up back in the adapter layer, so this concern is no longer relevant.

src/wayback/_http.py
```diff
@@ -0,0 +1,591 @@
...
+        return ''
+
+    @property
+    def is_success(self) -> bool:
+        """Whether the status code indicated success (2xx) or an error."""
+        return self.status_code >= 200 and self.status_code < 300
+
+    @property
+    def is_memento(self) -> bool:
+        return 'Memento-Datetime' in self.headers
```
Mr0grog, line 147: This property is not as generic as everything else in this class (which sticks to “HTTP” as the level of abstraction), but it was very helpfully convenient to add here. If we wanted to be more strict we could pull it back out into a utility function.

src/wayback/_http.py
```diff
@@ -0,0 +1,591 @@
...
+        Close the adapter and release any long-lived resources, like pooled
+        HTTP connections.
+        """
+        ...
+
+
+class RequestsAdapter(WaybackHttpAdapter):
+    """
+    Wrap the requests library for making HTTP requests.
+    """
```
Mr0grog, line 306: We now have two similarly named classes, which might be kind of dangerous:

- `RequestsAdapter` is private and just implements HTTP requests with the *requests* library. I wonder if it should have an underscore prefix or be renamed in some bigger way for clarity. (I *do* think keeping it separate is good, since it will get way more complex when we add thread safety in the next PR.)

- `WaybackRequestsAdapter` adds retry, rate limiting, etc. to the above class. It is the actual adapter that gets used. I’m also wondering if this should be renamed to something more generic or have a generic alias like `WaybackDefaultAdapter`. That way the public name is something people can rely on, so they don’t break if/when we swap out requests for raw urllib3 or httpx.

src/wayback/_http.py
```diff
@@ -0,0 +1,591 @@
...
+    # It seems Wayback sometimes produces 500 errors for transient issues, so
+    # they make sense to retry here. Usually not in other contexts, though.
+    retryable_statuses = frozenset((413, 421, 500, 502, 503, 504, 599))
+
+    # XXX: Some of these are requests-specific and should move WaybackRequestsAdapter.
+    retryable_errors = (ConnectTimeoutError, MaxRetryError, ReadTimeoutError,
+                        ProxyError, RetryError, Timeout)
+    # Handleable errors *may* be retryable, but need additional logic beyond
+    # just the error type. See `should_retry_error()`.
+    handleable_errors = (ConnectionError,) + retryable_errors
```
Mr0grog, line 404: These almost certainly need to move to either `RequestsAdapter` or `WaybackRequestsAdapter` in order for this to stay generic.

src/wayback/_http.py
```diff
@@ -0,0 +1,591 @@
...
+
+    # The implementation of different features is split up by method here, so
+    # `request()` calls down through a stack of overrides:
+    #
+    #   request                        (ensure valid types/values/etc.)
+    #   └> _request_redirectable       (handle redirects)
+    #      └> _request_retryable       (handle retries for errors)
+    #         └> _request_rate_limited (handle rate limiting/throttling)
+    #            └> request_raw        (handle actual HTTP)
+    def request(
```
Mr0grog, line 430: This breaks the class up and makes its definition longer, but hopefully it makes it a little more manageable than the way things were done in `WaybackSession`.

These could also be implemented as totally separate adapters, which might be cleaner? But it felt simpler to just keep them together for now.

This was also partly influenced by my realization that we have a bug where we don’t implement our own redirect following, which means redirects don’t get rate limited (sigh). It this gives us a good place to slide *that* complexity in later (`_request_redirectable()` is a pass-through in this PR).

src/wayback/_http.py
```diff
@@ -0,0 +1,591 @@
...
+                # The memento limit is actually a generic Wayback limit.
+                '/': RateLimit.make_limit(memento_rate_limit),
+            }
+        )
+        self.timeout = timeout
+        self.headers = {
+            'User-Agent': (user_agent or
+                           f'wayback/{__version__} (+https://github.com/edgi-govdata-archiving/wayback)'),
+            'Accept-Encoding': 'gzip, deflate'
+        }
```
Mr0grog, line 571: Now that I’m restructuring things, I wonder a bit whether the defaults for `user_agent` and the rate limits belong here, or in `WaybackClient`. As it is, someone wanting to provide a custom adapter needs to make sure they set the right defaults, which are very much tied to “correct” functioning of the overall library.

OTOH, keeping them here keeps configuration simple — you don’t really provide options to `WaybackClient` other than an instance of `WaybackHttpAdapter`, and all the various config options live on the adapter.

src/wayback/_http.py
```diff
@@ -0,0 +1,591 @@
...
+        method: str,
+        url: str,
+        *,
+        params: dict = None,
+        headers: dict = None,
+        follow_redirects: bool = True,
+        timeout: Union[int, Tuple[int, int]] = -1
+    ) -> WaybackHttpResponse:
+        timeout = self.timeout if timeout is -1 else timeout
+        headers = dict(headers or {}, **self.headers)
```
Mr0grog, line 584: Wondering if these somehow belong on the base `WaybackHttpAdapter` rather than here. 🤷

src/wayback/_utils.py
```diff
@@ -322,26 +320,25 @@ def __exit_all__(self, type, value, traceback):
         pass
 
 
-class DisableAfterCloseSession(requests.Session):
+class DisableAfterCloseAdapter:
```
Mr0grog, line 323: I wound up keeping this but changing the implementation slightly (so it mixes in kind of like `RetryAndRateLimitAdapter` does). It should probably move to `_http.py`.

src/wayback/exceptions.py
```diff
@@ -101,8 +101,8 @@ def __init__(self, response, retry_after):
         super().__init__(message)
 
 
-class SessionClosedError(Exception):
+class AlreadyClosedError(Exception):
     """
-    Raised when a Wayback session is used to make a request after it has been
+    Raised when a Wayback client is used to make a request after it has been
```
Mr0grog, line 106: It didn’t make sense to keep “session” in the name, but instead of replacing it with “adapter” I decided to go with making it a little more generic.

src/wayback/tests/test_client.py
```diff
@@ -755,13 +772,19 @@ def test_timeout_applied_request(self, mock_class):
...
         assert res.text == 'None'
         # Overwriting the default
         res = session.request('GET', 'http://test.com', timeout=1)
         assert res.text == '1'
 
+    @mock.patch('requests.Session.send', side_effect=return_user_agent)
+    def test_user_agent(self, mock_class):
+        adapter = WaybackRequestsAdapter()
+        agent = adapter.request('GET', 'http://test.com').text
+        assert agent.startswith('wayback/')
```
Mr0grog, line 786: I feel lucky I even noticed we needed this! We probably should have had a test before — the previous implementation was subtle because it used built-in functionality of `requests.Session` that changed when we stopped subclassing that.

Mr0grog (commented):
OK, I went ahead and moved the retry/rate limiting functionality down to the adapter layer, which effectively makes `WaybackHttpAdapter` a replacement for `WaybackSession`. We could keep the old name, but I think the new name is good and helps create distance from `requests.Session` (and also makes it clear that all the other Session functionality from `requests` that someone could have relied on is no longer there).

I also tried to cover my discomfort with all the retry/rate limit stuff being complex to customize an adapter around by making it into a sort of mixin:

```py
class RequestsAdapter(WaybackHttpAdapter):
    # Implements HTTP requests via the "requests" library.
    ...

class RetryAndRateLimitAdapter(WaybackHttpAdapter):
    # Mixin class that implements retrying and rate limiting
    # around the `adapter.request()` method.
    ...

class WaybackRequestsAdapter(RetryAndRateLimitAdapter, DisableAfterCloseAdapter, RequestsAdapter):
    # The actual adapter we want to use.
    ...
```

That’s a little complicated! But at least it means someone can leverage the existing retry and rate limiting functionality but plug in a different HTTP implementation. I *think* that’s good?

Anyway, there’s a lot of little cleanup stuff that has to happen here (also docs!!!!). But I think the overall shape of things is more-or-less what we want now.

PR #186, opened 2026-04-07
Update licensing info per PEP 639

### LINKED ISSUE #168: Update licensing info to conform with PEP 639
In `pyproject.toml`, we currently set the `project.license` field to a table and have a note about there being no Trove classifier for our license (3-Clause BSD): https://github.com/edgi-govdata-archiving/wayback/blob/47a436c27b83b3cb40bf18afcc12d3f3bc7b14a3/pyproject.toml#L13-L21

However, [PEP 639](https://peps.python.org/pep-0639/) is now finalized, and it updates the `project.license` field to be a SPDX identifier string. So we should change it to `"BSD-3-Clause"` and remove the comment about the Trove identifier. We *may* also want to set the `license-files` field (an array of paths/globs), but I _think_ it should get correctly set automatically since we use a pretty standard name (`LICENSE`). Needs checking.

MiTo0o:
Updated the project.license field to the SPDX identifier "BSD-3-Clause" per PEP 639 and removed the outdated Trove classifier comment. Verified locally by building the wheel that the build backend automatically includes the LICENSE file in the .dist-info directory, so an explicit license-files array is not needed. Closes #168

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -65,6 +65,7 @@ Thanks to the following people for their contributions and help on this package!
 
 - `Dan Allan <https://github.com/danielballan>`_ (Code, Tests, Documentation, Reviews)
 - `Rob Brackett <https://github.com/Mr0grog>`_ (Code, Tests, Documentation, Reviews)
+- `Derzan Chiang <https://github.com/MiTo0o>`_ (Documentation)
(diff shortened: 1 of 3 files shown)


Mr0grog (approved):
Thanks for fixing this up and checking on the actual build outputs here, too! Looks great. 🚀

PR #187, opened 2026-04-24
Rename CdxRecord fields to align with CDX server names (#130)

### LINKED ISSUE #130: Align CdxRecord attribute names with actual CDX names
Our `CdxRecord` objects have names that do not exactly align with the corresponding field names on the CDX server, or what you receive as column headers if asking for CDX search results in JSON format. We should allow access by those names in addition to the names we currently use (and maybe deprecate the current ones?).

- `key` should be `urlkey`
- `mime_type` should be `mimetype`
- `status_code` should be `statuscode`
- `url` should be `original`

MiTo0o:
Rename CdxRecord attributes to match the field names used by the CDX
server:
```
  key          -> urlkey
  url          -> original
  mime_type    -> mimetype
  status_code  -> statuscode
```

## Changes

- `CdxRecord` migrated from `collections.namedtuple` to
  `typing.NamedTuple` so the class can host the deprecation properties
  (and gets proper type hints for free).
- `raw_url` and `view_url` are now computed `@property`s derived from
  `original` and `timestamp`, per  #130.
- `_client.py` simplified — `search()` no longer does the
  `CdxRecord(*split, '', '')` + `_replace(...)` dance; it now just
  unpacks the 7 CDX fields and constructs the record in one step.
- The docstring moved inside the class body (it was a floating
  module-level string before).

### DIFF
--- a/src/wayback/_client.py
+++ b/src/wayback/_client.py
@@ -818,44 +818,38 @@ class WaybackClient(_utils.DepthCountedContext):
                     previous_result = text
 
                 try:
-                    data = CdxRecord(*text.split(' '), '', '')
-                    if data.status_code == '-':
-                        # the status code given for a revisit record
-                        status_code = None
-                    else:
-                        status_code = int(data.status_code)
-                    length = None if data.length == '-' else int(data.length)
-                    capture_time = _utils.parse_timestamp(data.timestamp)
+                    # The CDX server returns 7 fields:
+                    # urlkey, timestamp, original, mimetype, statuscode, digest, length
+                    (urlkey, raw_timestamp, original, mimetype,
+                     raw_status, digest, raw_length) = text.split(' ')
+
+                    statuscode = None if raw_status == '-' else int(raw_status)
+                    length = None if raw_length == '-' else int(raw_length)
+                    timestamp = _utils.parse_timestamp(raw_timestamp)
                 except Exception as err:
                     if 'RobotAccessControlException' in text:
                         raise BlockedByRobotsError(query["url"])
                     raise UnexpectedResponseFormat(
                         f'Could not parse CDX output: "{text}" (query: {sent_query})') from err
 
-                clean_url = REDUNDANT_HTTPS_PORT.sub(
+                original = REDUNDANT_HTTPS_PORT.sub(
                     r'\1\2', REDUNDANT_HTTP_PORT.sub(
-                        r'\1\2', data.url))
-                if skip_malformed_results and is_malformed_url(clean_url):
+                        r'\1\2', original))
+                if skip_malformed_results and is_malformed_url(original):
                     continue
-                if clean_url != data.url:
-                    data = data._replace(url=clean_url)
 
                 # TODO: repeat captures have a status code of `-` and a mime type
                 # of `warc/revisit`. These can only be resolved by requesting the
                 # content and following redirects. Maybe nice to do so
                 # automatically here.
-                data = data._replace(
-                    status_code=status_code,
-                    length=length,
-                    timestamp=capture_time,
-                    raw_url=_utils.format_memento_url(
-                        url=data.url,
-                        timestamp=data.timestamp,
-                        mode=Mode.original.value),
-                    view_url=_utils.format_memento_url(
-                        url=data.url,
-                        timestamp=data.timestamp,
-                        mode=Mode.view.value)
+                data = CdxRecord(
+                    urlkey=urlkey,
+                    timestamp=timestamp,
+                    original=original,
+                    mimetype=mimetype,
+                    statuscode=statuscode,
+                    digest=digest,
+                    length=length
                 )
                 count += 1
                 yield data
--- a/src/wayback/_client.py
+++ b/src/wayback/_client.py
@@ -15,7 +15,6 @@ Other potentially useful links:
 
 from base64 import b32encode
 from datetime import date
-from enum import Enum
 import hashlib
 import logging
 import re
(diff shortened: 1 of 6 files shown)


src/wayback/_client.py
```diff
@@ -818,44 +818,38 @@ def search(self, url, *, match_type=None, limit=1000, offset=None,
...
-                        # the status code given for a revisit record
-                        status_code = None
-                    else:
-                        status_code = int(data.status_code)
-                    length = None if data.length == '-' else int(data.length)
-                    capture_time = _utils.parse_timestamp(data.timestamp)
+                    # The CDX server returns 7 fields:
+                    # urlkey, timestamp, original, mimetype, statuscode, digest, length
+                    (urlkey, raw_timestamp, original, mimetype,
+                     raw_status, digest, raw_length) = text.split(' ')
```
Mr0grog: Can you add a splat at the end of this assignment? At some previous points in time (when they’ve been testing new things), archive.org’s CDX API has returned more fields. Without the splat, that would cause this code to raise.

```suggestion
                    (urlkey, raw_timestamp, original, mimetype,
                     raw_status, digest, raw_length, *_) = text.split(' ')
```

src/wayback/_models.py
```diff
@@ -1,77 +1,108 @@
...
 
-And these attributes are synthesized from the information provided by CDX.
+    .. py:attribute:: view_url
 
-.. py:attribute:: raw_url
+       The URL to the public view on Wayback Machine. In this view, the links and
+       some subresources in the document are rewritten to point to Wayback URLs.
+       There is also a navigation panel around the content. Example URL:
+       :data:`'https://web.archive.org/web/19961231235847/http://www.nasa.gov/'`.
+    """
```
Mr0grog: Now that this is using `class` syntax, does this work using [Numpydoc’s attribute syntax](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring) for the docstring instead of `.. py:attribute:: <name>`?

We generally use Numpydoc syntax in this package (and most EDGI Web Monitoring Python code). I *think* this one didn’t because it doesn’t work correctly with the `namedtuple` function. But hopefully it *should* work right with this new approach.

You can build the docs locally or preview them for this PR at: https://wayback--187.org.readthedocs.build/en/187/usage.html#wayback.CdxRecord We mainly want to make sure that they continue to get listed out as separate, describable items. Ideally they also work nicely with editor autocomplete/intellisense stuff (the old ones only sort of do).

src/wayback/_models.py
```diff
@@ -1,77 +1,108 @@
...
+
+    @property
+    def status_code(self) -> Optional[int]:
+        warn('The `status_code` attribute on `CdxRecord` was renamed to `statuscode`.',
+             DeprecationWarning, stacklevel=2)
+        return self.statuscode
+
+    @property
+    def raw_url(self) -> str:
+        # 'id_' = Mode.original.value (hardcoded to avoid circular import with _client.py)
```
Mr0grog: I think it should be OK to move the `Mode` enum to `_models.py`. I’m thinking I originally left it in `_client.py` because it was only used there, but it’s not really private, so it probably always should have been in `_models.py`… 😕

Mr0grog (changes requested):
This is great, thank you!

I left a few notes inline, mostly pretty minor stuff. The only serious one is the splat when parsing the CDX response.

Would you also please add a test that covers the `view_url` and `raw_url` properties of `CdxRecord`? I know we didn’t have anything for that before, but now that you are forcing me to look at it with fresh eyes, that seems like a problem, especially now that it is lazily computed as a property. (If you want to really make it nice, we should have a test that ensures all the resulting attributes were correctly parsed from a CDX line in `WaybackClient.search()`.)

Mr0grog:
> `_client.py` simplified — `search()` no longer does the
>   `CdxRecord(*split, '', '')` + `_replace(...)` dance; it now just
>   unpacks the 7 CDX fields and constructs the record in one step.

❤️

src/wayback/_models.py
MiTo0o: ah this is great news, I was trying to figure out how to get this to work without circular importing haha

MiTo0o:
Sorry this took a while, life caught up for a bit. Let me know if the [docs](https://wayback--187.org.readthedocs.build/en/187/usage.html#wayback.CdxRecord) for the attributes look ok to you? 
It sort of bothers me that this line:
`And these attributes are synthesized from the information provided by CDX.`
renders with grey highlighting(sphinx might think its an attribute header), but I'm not sure how to fix this.

Mr0grog (approved):
This is great! 👍 

I made a bunch of tweaks, but mainly nothing that was a problem in what you did — mostly just cleaning up *my own* mess.

> Sorry this took a while, life caught up for a bit.

No worries!

> It sort of bothers me that this line: `And these attributes are synthesized from the information provided by CDX.` renders with grey highlighting(sphinx might think its an attribute header)

Yeah, I think that’s what’s happening.

I went ahead and just removed this line — I don’t think it and the earlier matching line (`These attributes contain information provided directly by CDX.`) ever actually made much sense. They are not very meaningful (possibly even confusing) for most users of this package, who are probably not deeply familiar with the CDX API.

> We generally use Numpydoc syntax in this package (and most EDGI Web Monitoring Python code). I think this one didn’t because it doesn’t work correctly with the `namedtuple` function.

Well, I feel like an idiot. This was totally wrong on my part, and I even left myself a comment just lower down in the same file about it! 🤦 

https://github.com/edgi-govdata-archiving/wayback/blob/1b8a19142005241764a3b589b153a0c9b64f8269/src/wayback/_models.py#L77-L81

I’ve switched this back to `.. py:attribute::` syntax for now; these issues are their own little mess that should stay separate from this PR, and may require upstream fixes anyway. Sorry I sent you off in the wrong direction on this.

(I am also realizing I never replied to the Numpydoc maintainers on that issue, which is not great. I guess this is a good reminder for me! 😅)

PR #192, opened 2026-04-30
Give Memento class a nicer repr

### LINKED ISSUE #105: Give Memento objects a nicer repr
Calling `repr()` with a `Memento` object gets you a not-very-useful representation:

```
<wayback._models.Memento object at 0x102447970>
```

Instead, this should include some more info (at least the URL and timestamp?) and not include `_models` since users shouldn’t be importing that.

For example:

```
<wayback.Memento url="https://www3.epa.gov/" timestamp="20221001000000">
<wayback.Memento url="https://www3.epa.gov/" timestamp="2022-10-01T00:00:00Z">
<wayback.Memento "https://www3.epa.gov/" at 2022-10-01T00:00:00Z>
```

Or something along those lines.

This should just involve adding a `__repr__(self)` method to [`Memento`](https://github.com/edgi-govdata-archiving/wayback/blob/0b2134b92d6eb66c2eeaa0f465c692b1d88ffdf8/wayback/_models.py#L77-L257).

MiTo0o:
Simple changes for #105, nicer `__repr__` for Memento Class

Example:
```py
from datetime import datetime, timezone
from wayback._models import Memento, Mode                                                   
m = Memento(
    url='https://www3.epa.gov/',
    timestamp=datetime(2022, 10, 1, 0, 0, 0, tzinfo=timezone.utc),
    mode='id_',
    memento_url='https://web.archive.org/web/20221001000000id_/https://www3.epa.gov/',
    status_code=200,
    headers={},
    encoding='utf-8',
    raw=None,
    raw_headers={},
    links={},
    history=[],
    debug_history=[]
)
repr(m)
```

Outputs:
`'<wayback.Memento url="https://www3.epa.gov/" timestamp="2022-10-01T00:00:00+00:00">'`

### DIFF
--- a/src/wayback/_models.py
+++ b/src/wayback/_models.py
@@ -434,6 +434,11 @@ class Memento:
     def __exit__(self, *_args):
         self.close()
 
+    def __repr__(self):
+        return (f'<{type(self).__module__.rsplit(".", 1)[0]}'
+                f'.{type(self).__name__} url="{self.url}" '
+                f'timestamp="{self.timestamp.isoformat()}">')
+
     @classmethod
     def parse_memento_headers(cls, raw_headers, url='https://web.archive.org/'):
         """
--- a/src/wayback/_models.py
+++ b/src/wayback/_models.py
@@ -435,7 +435,7 @@ class Memento:
         self.close()
 
     def __repr__(self):
-        return (f'<{type(self).__module__.rsplit(".", 1)[0]}'
+        return (f'<{type(self).__module__.split("._", 1)[0]}'
                 f'.{type(self).__name__} url="{self.url}" '
                 f'timestamp="{self.timestamp.isoformat()}">')
 
--- a/src/wayback/tests/test_models.py
+++ b/src/wayback/tests/test_models.py
@@ -54,5 +54,4 @@ def test_memento_repr():
         history=[],
         debug_history=[]
     )
-    print(repr(memento))
     assert repr(memento) == '<wayback.Memento url="https://www3.epa.gov/" timestamp="2022-10-01T00:00:00+00:00">'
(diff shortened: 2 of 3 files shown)


src/wayback/tests/test_models.py
```diff
@@ -37,3 +37,22 @@ def test_cdx_record_deprecated_fields():
...
+        status_code=200,
+        headers={},
+        encoding='utf-8',
+        raw=None,
+        raw_headers={},
+        links={},
+        history=[],
+        debug_history=[]
+    )
+    print(repr(memento))
```
Mr0grog: Please remove this!

src/wayback/_models.py
```diff
@@ -450,6 +434,11 @@ def __enter__(self):
     def __exit__(self, *_args):
         self.close()
 
+    def __repr__(self):
+        return (f'<{type(self).__module__.rsplit(".", 1)[0]}'
```
Mr0grog: Oh, neat, doing this dynamically allows possible inheriting classes to leverage it!

But I think this could be slightly better (more abstract) if you `split("._", 1)[0]`, so it’s just cutting out “private” modules:


```suggestion
        return (f'<{type(self).__module__.split("._", 1)[0]}'
```

Or make it just work for this package (a little more complicated, maybe not worthwhile):

```py
        module = type(self).__module__
        return (f'<{module.split(".", 1)[0] if module.startswith(__name__) else module}'
```

Mr0grog (changes requested):
This is great, thanks! Please clean up the `print()` statements, though. 😉 

One other minor note inline, not as big a deal.

src/wayback/tests/test_models.py
MiTo0o: OH yikes, i forgot!

src/wayback/_models.py
MiTo0o: my vote is for repr to stay simple/straightforward :D I like your first suggestion more

Mr0grog (approved):
Perfect, merging this in! 👏 🙇

PR #193, opened 2026-04-30
Add support for `datetime.timedelta` to `get_memento()`

### LINKED ISSUE #55: Support `datetime.timedelta` for `target_window` in `get_memento()`
The `WaybackClient.get_memento()` method has a `target_window` parameter, which is used when a memento does not exist at the requested time. If the parameter `exact=False`, `get_memento()` returns the nearest-in-time memento, but only if it is within the number of seconds specified in `target_window`.

https://github.com/edgi-govdata-archiving/wayback/blob/c066e04c1ba11c26febac574b37ee739cc5796e2/wayback/_client.py#L717-L719

Since Python has a built in `datetime.timedelta` type to represent exactly this kind of concept, we should also accept it instead of an integer for `target_window`. (Really, I’m not sure why we didn’t do this originally!)

MiTo0o:
I'm assuming you wanted to support both seconds and timedelta per #55?

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -5,9 +5,6 @@ Release History
 (In Development)
 ----------------
 
-- Updated license identifier to follow PEP 639 style. This should make sure the license is surfaced better and more clearly on PyPI. (:issue:`186`)
-
-
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


Mr0grog (approved):
> I'm assuming you wanted to support both seconds and timedelta per #55?

Yes! Sorry if that was poorly worded. This is perfect.

PR #194, opened 2026-05-04
test: disable rate limiting in tests using shared test_client fixture

### LINKED ISSUE #190: Turn down rate limiting when using VCR cassettes and mock responses in tests
I could swear we used to do turn off rate limiting in tests,, but while reviewing #187, I realized the test suite runs very slowly because it is rate limiting even for tests that are not specifically of rate-limiting functionality.

The simplest solution here is probably to have a pytest fixture that provides a pre-configured `WaybackClient` instance, rather than doing `with WaybackClient()` everywhere. For example, instead of this test: https://github.com/edgi-govdata-archiving/wayback/blob/195add2a6ed7e6bac9dc7dce45c8e2b73a0ce274/src/wayback/tests/test_client.py#L65-L73

…we could have:

```py
@ia_vcr.use_cassette()
def test_search_with_date(test_client):
    versions = test_client.search('dw.com',
                                  from_date=date(2019, 10, 1),
                                  to_date=date(2020, 3, 1))
    for v in versions:
        assert v.timestamp >= datetime(2019, 10, 1, tzinfo=timezone.utc)
        assert v.timestamp <= datetime(2020, 3, 1, tzinfo=timezone.utc)
```

Another alternative might be to have a test decorator that monkey-patches the global default rate limits during the test function. Lots of possible approaches here.

Adeelp1:
fix: #190 

### Summary

Disable rate limiting in tests that use VCR cassettes and mocked responses by introducing a shared pytest fixture.

### Changes

- Added a test_client fixture in conftest.py with rate limits disabled
- Updated tests to use the shared client when using VCR (@ia_vcr.use_cassette()) and requests_mock
- Left rate-limit-specific tests unchanged to preserve existing behavior

### Result

- Faster test execution
- Cleaner and more consistent test setup

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,6 +63,7 @@ Contributors
 
 Thanks to the following people for their contributions and help on this package! See our `contributing guidelines <https://github.com/edgi-govdata-archiving/wayback/blob/main/CONTRIBUTING.rst>`_ to find out how you can help.
 
+- `@Adeelp1 <https://github.com/Adeelp1>`_ (Tests)
 - `Dan Allan <https://github.com/danielballan>`_ (Code, Tests, Documentation, Reviews)
 - `Rob Brackett <https://github.com/Mr0grog>`_ (Code, Tests, Documentation, Reviews)
 - `Derzan Chiang <https://github.com/MiTo0o>`_ (Code, Tests, Documentation)
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -61,12 +61,15 @@ Fixes & Maintenance
 
 - :class:`wayback.Memento` now has a nicer, more informative ``repr`` when you are using a notebook or Python REPL. (:issue:`192`)
 
+- Tests where rate limits don’t matter now run faster. (:issue:`194`)
+
 
 New Contributors
 ^^^^^^^^^^^^^^^^
 
-- `Derzan Chiang <https://github.com/MiTo0o>`_ (Code, Tests, Documentation)
-- `Beckett Frey <https://github.com/BeckettFrey>`_ (Code, Tests, Documentation)
+- `Derzan Chiang <https://github.com/MiTo0o>`_
+- `Beckett Frey <https://github.com/BeckettFrey>`_
+- `@Adeelp1 <https://github.com/Adeelp1>`_
 
 
 v0.4.5 (2024-02-01)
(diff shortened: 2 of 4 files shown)


Adeelp1:
Note: This test (`test_get_memento_with_archive_url`) was already failing locally due to newline differences (`\r\n` vs `\n`). This is unrelated to the rate-limiting changes.

Mr0grog (approved):
@Adeelp1 sorry I haven’t had time to get back to you this week. Looks great! 👍 

> Note: This test (`test_get_memento_with_archive_url`) was already failing locally due to newline differences (`\r\n` vs `\n`). This is unrelated to the rate-limiting changes.

I’m not sure I’m following. Is this test failing locally for you? It seems to work fine for me and on CI. Are you on Windows? That *shouldn’t* affect the `content == memento.content` line since it’s just looking at the binary representation, but I wonder if something in VCR or [our wacky serializer hack](https://github.com/edgi-govdata-archiving/wayback/blob/5cd58a23267e451a02e6cf0d463d70b1b2daa548/src/wayback/tests/support.py#L61-L124) is causing problems. The bytes in the file and in the cassette YAML are the same. 😕

Which representation are you seeing `\r\n` vs just `\n` in?

Adeelp1:
Yes, I’m using Windows. I’m seeing a newline mismatch locally:

* `content` -> `\r\n`
* `memento.content` -> `\n`

Mr0grog:
Ahhhhhh, I do so little development on Windows these days, I had totally forgotten about this, but it’s a _git_ problem. On Windows, git will automatically convert `\n` line endings to `\r\n` when checking out files (and the opposite when committing). So when you check out a commit, git is messing with the contents of `src/wayback/tests/test_files/fws-gov-birds.txt`, which breaks the test. At least that is what was happening for me when I was able to reproduce your problem just now.

I think we can fix that for these test files by marking them as binary for git. But this is all clearly unrelated to this particular PR. Can you file a separate issue to track it, please?

Mr0grog:
> Can you file a separate issue to track it, please?

Nevermind, I went ahead and filed an issue for it in #197.

Mr0grog (approved):
Alright, I’m going to go ahead and merge this.

@Adeelp1, please let me know on #199 if that helps solve your local testing issues.

PR #199, opened 2026-05-12
Tell git not to mess with test data on Windows

### LINKED ISSUE #197: Test `test_get_memento_with_archive_url` is broken on Windows
For most developers on Windows, the `test_get_memento_with_archive_url` test will be broken: https://github.com/edgi-govdata-archiving/wayback/blob/3a0dbf3156b2cdb7c9a172382e5318ebb44a923f/src/wayback/tests/test_client.py#L443-L469

Specifically, this assertion fails:

```py
content = get_file('fws-gov-birds.txt')
assert content == memento.content
```

This isn’t really a problem with test. It turns out git on Windows messes with the line endings of files when checking them out and committing them back in (it converts `\n` line endings to `\r\n`, which are how Windows generally does line endings). That’s convenient a lot of the time, but this test needs to check byte-for-byte equality, so it’s a problem (when git checks out `fws-gov-birds.txt`, it changes the file’s bytes).

We *should* be able to fix this for Windows devs by marking these test files as binary files in `.gitattributes`.

Mr0grog:
On Windows, git will automatically change the line endings on text files. Usually that’s a useful convenience, but our test data needs to have exact bytes that do not change. This marks those files as non-text so git doesn’t mess with them and cause testing issues. (We use non-text instead of binary so that we can still get nice diffs!)

Fixes #197.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,4 +2,4 @@
 *.py diff=python
 
 # Treat test data as binary so git does not change line endings or other bytes.
-src/wayback/tests/test_files/* binary
+src/wayback/tests/test_files/* -text
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,9 @@
 .git_archival.txt  export-subst
 *.py diff=python
 
-# Don't treat test data text so git does not change line endings or other bytes.
+# Don't treat test data as text. Git will change some features of text files
+# when checking out and committing (e.g. line endings on Windows), but the test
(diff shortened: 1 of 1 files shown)


Mr0grog:
@Adeelp1 does this branch work for you? You may need to delete the `test_files` directory and then `git restore` it.

Adeelp1:
@Mr0grog Yes, this works for me on Windows.

PR #201, opened 2026-05-13
Cache HTTP responses for docs with VCR

### LINKED ISSUE #188: Mock HTTP responses in docs build or remove ipython
While reviewing #187, I was reminded of how our docs build can be slow and failure-prone because we use iPython to evaluate the example code, and that means building docs makes real requests to the Wayback Machine. Not great.

Example: https://github.com/edgi-govdata-archiving/wayback/blob/195add2a6ed7e6bac9dc7dce45c8e2b73a0ce274/docs/source/usage.rst?plain=1#L21-L26

Ideally, we can set up VCR in the docs examples [using `ipython_execlines`](https://ikerdocs-sphinx.readthedocs.io/extras/IPython.html) or similar, similar to how we use VCR in tests. That way, the examples don’t actually go out and make live requests of the Wayback Machine.

If there’s no workable or reasonable way to do that, we should switch all those example code blocks off iPython (maybe just remove iPython altogether), since I think it’s probably causing for harm and instability than good right now.

Mr0grog:
When a lot of builds happen at once, our docs frequently start to fail because of rate limits with the Wayback Machine. Docs builds are also often slow. Both of these are because we use ipython to execute and show the results of code samples in the tutorial, which means we hit live Wayback Machine APIs. This solves the problem by recording and replaying the API responses using VCR, just like we do for tests.

I couldn’t figure out a way to do this entirely with the `ipython_execlines` setting, since it doesn’t seem to give you a way to execute cleanup/teardown code, and VCR needs that to write out its results. So I wound up with a kind of weird hybrid solution, where I use `ipython_execlines` to set a directory path to read/write VCR cassettes in, and then have suppressed (hidden) ipython directives at the start and end of the tutorial to begin and clean up VCR recording and replay. It’s not the prettiest approach, but it works.

Fixes #188.

### DIFF
--- a/docs/source/cassettes/tutorial.yaml
+++ b/docs/source/cassettes/tutorial.yaml
@@ -1,0 +1,3220 @@
+interactions:
+- request:
+    body: null
+    headers:
+      Accept-Encoding:
+      - gzip, deflate
+      User-Agent:
+      - wayback/0.4.5.dev32+ge28b94a3d.d20260512 (+https://github.com/edgi-govdata-archiving/wayback)
+    method: GET
+    uri: https://web.archive.org/cdx/search/cdx?url=nasa.gov&limit=1000&showResumeKey=true&resolveRevisits=true
+  response:
+    body:
+      string: 'gov,nasa)/ 19961231235847 http://www.nasa.gov:80/ text/html 200 MGIGF4GRGGF5GKV6VNCBAXOE3OR5BTZC
+        1811
+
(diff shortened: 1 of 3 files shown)


Mr0grog:
Rebased to fix merge conflicts.

danielballan:
Yes, there is no teardown hook in the `ipython` sphinx directive. You could hook into sphinx's own teardown like so, but I think staying within the `ipython` sphinx directives own system is more discoverable/explicit.

```python
def cleanup(app, exception):
    ...

def setup(app):
    app.connect("build-finished", cleanup)
```

Mr0grog:
@danielballan makes sense, thanks for taking a look!

PR #216, opened 2026-07-08
Retry transient connection resets in WaybackSession

soodoku:
## What this does

While running a large research crawl (a few thousand mementos + CDX searches at the default rate limits), requests intermittently failed with:

```
requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(54, 'Connection reset by peer'))
```

These were raised immediately rather than retried, because `should_retry_error()` only string-matched for `'NewConnectionError'` and `'Max retries'`. Server-side disconnections like this are transient (they appear when Wayback Machine servers shed load), so this PR treats them as retryable — including the `RemoteDisconnected` case the existing `NOTE` comment in that method already flagged as probably warranting a retry (`http.client.RemoteDisconnected` subclasses `ConnectionResetError`, so it comes along for free).

## Why type-checking instead of another string match

`str(error)` is not a reliable signal for this family of errors: depending on how the underlying error is nested, OSError's errno-style formatting can drop the wrapped exception's class name entirely, e.g.

```
>>> str(requests.exceptions.ConnectionError('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer')))
'[Errno Connection aborted.] [Errno 104] Connection reset by peer'
```

So `should_retry_error()` now walks the chain of wrapped exceptions (`__cause__`/`__context__`, via a new `_utils.iterate_wrapped_exceptions()`) and checks the underlying types. Both requests and urllib3 wrap by raising inside an `except` block, so `__context__` reliably links each layer; `MaxRetryError` is raised with explicit chaining (`raise MaxRetryError(...) from reason`), so `__cause__` covers that one.

The `'NewConnectionError'`/`'Max retries'` string checks are replaced by an `isinstance` check for `urllib3.exceptions.NewConnectionError` on that same chain, so there is no string matching left in `should_retry_error()`.

## Testing

- New `test_retries_connection_reset` (end-to-end via `requests_mock`: reset then success).
- New `test_should_retry_error_on_transient_disconnects` covering a reset, a `RemoteDisconnected`, a refused connection (`MaxRetryError` → `NewConnectionError`), and a non-retryable negative case. A `chain_exceptions()` test helper builds these by raising each layer inside the previous one's `except` block, so the test exceptions have the same shape as the real ones.
- Verified the fix against a real socket that sends an RST, not just mocks.
- Full `test_client.py` suite passes (53 passed); `ruff check`, `ruff format --check`, and `mypy` are clean under the project config.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/wayback/_client.py
```diff
@@ -276,6 +277,26 @@ def _new_header_init(self, headers=None, **kwargs):
...
+    while stack:
+        current = stack.pop()
+        if id(current) in seen:
+            continue
+        seen.add(id(current))
+        yield current
+        stack.extend(arg for arg in getattr(current, 'args', ()) if isinstance(arg, BaseException))
+        for linked in (current.__cause__, current.__context__):
+            if linked is not None:
+                stack.append(linked)
```
Mr0grog: Please move this to `_utils.py`. Our tendency to leave utilities in this module if they aren’t used elsewhere has led to this module getting pretty cluttered and hard to navigate.

I would also like to avoid looking at `.args` here — like string checking, it leads to problems where it’s not clear whether we’re actually talking about a meaningful causing exception or something else (in fact, we currently do string checking in part to actively avoid looking at `.args`). Looking through Requests’s source, checking `__context__` will always get us what we are looking for when it does wrapping (Requests does not use explicit chaining which would set `__cause__`, though it would still be good to check in case that changes in the future).

If you drop the `.args` checking, I think you can also drop all the complexity with the `seen` set. The case where you might see the same instance across multiple nested `__cause__` and `__context__` values is going to be *vanishingly* rare, and not worth the overhead.

src/wayback/_client.py
```diff
@@ -430,6 +451,7 @@ def send(self, request: requests.PreparedRequest, **kwargs):
                     retry_delay = self.get_retry_delay(retries, response)
                     logger.info('Caught exception during request, will retry: %s', error)
                 else:
+                    logger.info('Caught non-retryable exception during request: %s', error)
```
Mr0grog: Please remove this. We are already raising an exception here, and it should absolutely be up to calling or user code to decide whether and how to log or swallow that information.

src/wayback/_client.py
```diff
@@ -461,13 +483,19 @@ def should_retry_error(self, error):
...
+            # more detailed errors from urllib3 (which in turn wrap the
+            # original OSError), so unwrap the chain and check the underlying
+            # types. `ConnectionResetError` and `RemoteDisconnected` show up
+            # when the server abruptly closes the connection (e.g. when
+            # Wayback sheds load); both are transient and worth retrying with
+            # backoff. Note `str(error)` is not a reliable signal here: for
+            # some argument shapes OSError formatting drops the wrapped
+            # exception's class name entirely.
+            for wrapped in _iterate_wrapped_exceptions(error):
+                if isinstance(wrapped, (ConnectionResetError, RemoteDisconnected)):
```
Mr0grog: Checking `RemoteDisconnected` is redundant here; it’s a subclass of `ConnectionResetError` so will already be covered by that check.

Could you also please move the `NewConnectionError` check here and remove the string checking altogether? If we’re going going to be unwrapping, we should do an actual `isinstance` check instead of looking for the string.

src/wayback/tests/test_client.py
```diff
@@ -707,6 +709,50 @@ def test_only_retries_some_errors(self, requests_mock):
...
+        )
+        requests_mock.get('http://test.com', [{'exc': reset_error}, {'text': 'good', 'status_code': 200}])
+        session = WaybackSession(retries=1, backoff=0.1)
+        response = session.request('GET', 'http://test.com')
+        assert response.status_code == 200
+
+    def test_should_retry_error_on_transient_disconnects(self):
+        session = WaybackSession(retries=1, backoff=0.1)
+        # The shape urllib3 actually produces: the original OSError nested
+        # inside a ProtocolError inside requests's ConnectionError.
```
Mr0grog: Is this comment actually pertinent? The code below should either reproduce exactly the error we expect to see (in which case this comment is redundant) or explain *why* there is a discrepancy from what we actually expect (if it only tells me the test is incorrect, that just adds confusion!).

And if the discrepancy is minor and not relevant, just say that, or don’t mention it at all.

Mr0grog (changes requested):

src/wayback/_client.py
soodoku: Moved to `_utils.py` as `iterate_wrapped_exceptions()`, and dropped the `.args` walk.

I checked this against real socket errors rather than assuming: for a server-side RST, requests raises `ConnectionError(ProtocolError("Connection aborted.", ConnectionResetError(...)))`, and every layer is also linked by `__context__`, because both requests and urllib3 wrap by raising inside an `except` block (`requests/adapters.py` `except (ProtocolError, OSError) as err: raise ConnectionError(err, ...)`, and urllib3 re-raising the `ProtocolError` it built while still handling the original `OSError`). For a refused connection, `MaxRetryError` is raised with explicit chaining (`raise MaxRetryError(_pool, url, reason) from reason`), so `__cause__` covers that. So `__context__`/`__cause__` alone reaches everything the `.args` walk did.

With `.args` gone the chain is linear, so the `seen` set went away too — it is now just a `while current is not None: yield current; current = current.__cause__ or current.__context__` loop.

src/wayback/_client.py
soodoku: Removed.

src/wayback/_client.py
soodoku: Both done: the `RemoteDisconnected` check is gone (it does subclass `ConnectionResetError`), and the string checks are replaced by an `isinstance` check for `urllib3.exceptions.NewConnectionError` on the unwrapped chain.

One note on dropping the `'Max retries'` check specifically, since that is a slightly wider net than `NewConnectionError`: looking at `requests/adapters.py`, a `MaxRetryError` only reaches the `raise ConnectionError(e, request=request)` branch after the `ConnectTimeoutError` / `ResponseError` / `ProxyError` / `SSLError` reasons have been peeled off into their own exception types (all of which are already in `retryable_errors`). `NameResolutionError` subclasses `NewConnectionError`, so that is covered too. So this should not lose retry coverage in practice.

src/wayback/tests/test_client.py
soodoku: Not pertinent — dropped it. The tests now build their exceptions by actually raising them in nested `except` blocks (via a `chain_exceptions()` helper), so they have the same `__context__` chain as the real errors instead of an approximation that needed a comment to explain the gap.

